PR #92, opened 2015-01-23
Fix for string slicing when expecting a list.

sjdines:
When in the Emails model administration: if a single email is listed in the 'to' field and a Log inline is created with an error in it the email is treated as a string but the _format_value function expects a list causing the email address to be sliced into a character list.

The fix proposes that the _format_value function checks if it is a basestring first and if so wraps it in a list.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
   - sudo ln -s /run/shm /dev/shm
 
 install:
-  - pip install Django==$DJANGO jsonfield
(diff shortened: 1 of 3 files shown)


selwin:
Thanks for the PR. Can you change the code to use django.utils.six ? This way we don't require people to download a fresh copy of six just for this check.

sjdines:
@selwin thanks for that call; updated now

selwin:
Merged, thanks!

PR #93, opened 2015-02-16
Fix migration error for 0007

jamescw:


selwin:
I would think that the default empty string would be more space efficient than an empty JSON object. What's the rationale behind this change?

jamescw:
The empty string kept causing the migration to fail due to the Json decoder throwing an exception, I will paste the stack trace in the morning but I'm surprised no one else has run into this. This indeed did fix the issue and migrations ran without trouble.

Sent from Outlook




On Mon, Feb 16, 2015 at 5:19 PM -0800, "Selwin Ong" <notifications@github.com> wrote:












I would think that the default empty string would be more space efficient than an empty JSON object. What's the rationale behind this change?

selwin:
Was the error thrown by the DB or Python? I personally use this library in production and ran into no such issue.

selwin:
Closing as there's no re

PR #103, opened 2015-05-29
Internationalization

jrief:
This adds support for internationalization to post_office.

Additionally I fixed a small bug in the template validation: a3285fd


selwin:
Thanks for the PR!

How would an internationalized template work with emails sent with `render_on_delivery=True`? The content of these emails are not rendered when they are first created so we'll need to store the language information with the email.

Additionally, when `send_queued_mail` is called, we'll incur an additional database hit for each email with language set.

I wonder if we should just add a `language` variable in `EmailTemplate` model and we'll simply assign the right template to an `Email` instance when it's first created. So we'd use it as such:
```python
template = EmailTemplate.objects.create(name=my_template, subject=subject, language=my_language)
mail.send('alice@example.com', 'bob@example.com', template=my_template, language=my_language)
```
What do you think?

jrief:
@selwin This PR was just a prove of concept which works fine for prerendered Emails.
 
I now added a field ``language`` to model ``Email``. This can be queried when rendering templates using the option ``render_on_delivery=True``. BTW, I couldn't find the function which renders the Email on delivery; any hint please.

selwin:
Sorry for the late reply. I wonder if we can implement this feature just by adding `language` and `default_template` columns on `EmailTemplate` model.

```python
class EmailTemplate(models.Model):
    # ...
    language = models.CharField(blank=True)
    default_template = models.ForeignKey(self, related_name='translated_templates')
```

This is how we would use it:
```python
template = EmailTemplate.objects.create(name='my_template', ...)
template.translated_templates.create(language=my_language)

mail.send(template=my_template, language=my_language)
```

`mail.send` would look at the `language` argument and simply assign the right `email_template`. This way, we won't have to change other parts of the library.

What do you think?

jrief:
Yes, this would be a DRYer approach, because we then would not have to duplicate the fields ``subject``, ``message`` and ``html_message``. But even the current version has some kind of field duplication: ``subject``, ``message`` and ``html_message`` are part of model ``Email`` and ``EmailTemplate``. Maybe an abstract base model should hold these 3 fields.

The reason I implemented it this way, was that a lot of users, especially in the US, do not and never will need internationalization. If we add a default template, we always have one level of indirection.
Additionally it makes migrations from the previous version much more difficult, because we would have to create a datamigration.

BTW: I already added internationalization to my current project and it works quite fine. I now use ``render_on_delivery=True`` which is really nice, because this speeds up the request-response-cycle even more.

jrief:
I just had a look at the code on how to rewrite class ``EmailTemplate``, so that we have a ``TranslatedEmailTemplate`` which also handles the default/fallback language.
A problem here is, that in ``EmailTemplateAdmin`` we then need at least one ``TranslatedEmailTemplateInline``, but until Django-1.6 you can't specify a minimum for this; the parameter [min_num](https://docs.djangoproject.com/en/1.7/ref/contrib/admin/#django.contrib.admin.InlineModelAdmin.min_num) for a ``InlineModelAdmin`` was recently introduced in Django-1.7.

I added one small change, so that the ``AdminInline`` for model ``TranslatedEmailTemplate`` is hidden in case ``settings.USE_I18N`` is ``False``. A large user base never needs internationalization and so we don't bother them with extra admin inlines. I hope this substantiation is OK for you.

Any news about this PR? I hope to see it soon in a new release of **django-post_office**!

jrief:
This PR now is waiting for one month and still neither commented nor merged.
I also believe that internationalization is a feature which is required by a big audience.

domtes:
+1

selwin:
@jrief sorry about the delay. I've been really busy and although I do try to develop features, it's mostly on features that we actually use.

I'd like to get this feature in, but I think I prefer the approach that I described as I think it will require much less code and more performant. Users who don't require internationalization won't have an extra layer of indirection as they'd just keep using it as it is now. `EmailTemplate`s that are created without the `language` field will be the default.

Again, I apologize for the delay!

jrief:
Now email content (subject, text, html) has been moved into its own abstract base class, which is shared by ``EmailTemplate`` and ``TranslatedEmailTemplate``.
First I tried to keep all these fields inside ``TranslatedEmailTemplate``. The big problem then however is, that its impossible to force the administrator to create some default content. Even, if you add a *default* ``TranslatedEmailTemplate``, you have no guarantee that its not deleted. The validation functions, such as ``clean()`` can't perform these checks.

Therefore IHMO the only safe solution, if you insist on a fallback template, is to store the fields ``subject``, ``text`` and ``html`` twice. Everything else does more harm than benefit.

selwin:
Thanks for the update.

Could you please get rid of the `TranslatedEmailTemplate` altogether? Sorry if my previous comment wasn't clear. It will be much easier if we just add `language` and `default_template` fields to `EmailTemplate`. So the model would look like this:
```python
class EmailTemplate(models.Model):
    """
    Model to hold template information from db
    """
    name = models.CharField(max_length=255, help_text=("e.g: 'welcome_email'"))
    description = models.TextField(blank=True, help_text='Description of this template.')
    subject = models.CharField(max_length=255, blank=True, validators=[validate_template_syntax])
    content = models.TextField(blank=True, validators=[validate_template_syntax])
    html_content = models.TextField(blank=True, validators=[validate_template_syntax])
    created = models.DateTimeField(auto_now_add=True)
    last_updated = models.DateTimeField(auto_now=True)
    language = models.CharField(blank=True, max_length=16)
    default_template = models.ForeignKey(self, related_name='translated_templates')
```

With this model, we don't event need to change much of the email sending related code/logic. We only need to patch `mail.send` to assign the right template based on the provided `language` argument.

jrief:
But this doesn't make any sense and it causes a lot of data duplication.
Say, an application wants to send some content. The application want to use a "logical" template. ``EmailTemplate`` is such a logical template offering the base content translated into different languages stored inside ``TranslatedEmailTemplate``.

Otherwise an administrator has to duplicate the entries for the fields ``name`` and ``description`` for each translated ``EmailTemplate``. It also gets much more difficult to group translated content into one logical template because we then loose all the hierarchy. And worst of all, you can't assign a logical template to an application wishing to send certain content.

@domtes since you upvoted for this feature, how is your usecase.

selwin:
I don't think there's more duplication than the proposed `TranslatedEmailTemplate` approach, the data being stored are the same after all (subject, content, HTML content).

My proposed API for adding an internationalized template I described above is (so the admin doesn't need to re-type the name and description):
```python
template = EmailTemplate.objects.create(name='my_template', ...)
template.translated_templates.create(language=my_language, # subject, content...)
```
Obviously, translated templates can also be created via admin inlines.

To get all translations for a template, you can do `template.translated_templates.all()` so there's definitely hierarchy there :)

selwin:
And plus, the *really* nice thing about this is that we don't need to touch the `Email` model at all, since the translated email template is just a `Template` object. We'll be able to save a bit of space this way, since we won't need to store `language` field on the `Email` model which typically have many, many more rows than `EmailTemplate`.

jrief:
Yes, by having ``EmailTemplate `` refer to itself, we can avoid an extra table at the cost of having fields ``name`` and ``description`` which then often are empty. I implemented it this way now.

selwin:
Superseded by #105

PR #105, opened 2015-07-03
Internationalization - redux

selwin:
@jrief I made some changes to simplify things a bit more. `EmailTemplate` is now assigned to `Email` instances in `mail.send()`. This allows us to:

- Keep email rendering logic simple

- Avoid the use of `prefetch_related` when sending queued emails. We often send tens of thousands of emails in one batch, using `prefetch_related` would significantly increase memory usage in this case because each email will have it's own copy of all translated templates

- Additionally, it will now fail loudly if you try to assign a non existing language instead of falling silently to the default template.

Comments?


jrief:
Great that we finally get i18n support in post_office.

> Additionally, it will now fail loudly if you try to assign a non existing language instead of falling silently to the default template.

I implemented this intentionally. Reason is, that sometimes the language is taken directly from the request object. Then a fallback might be a better solution, but I don't insist on this, it was just a consideration. Please implement it the way you feel most comfortable.

jrief:
Thanks for merging. I just updated to the latest German translations.

PR #107, opened 2015-07-15
Fix default translated template

jrief:
Unfortunately my former PR had a bug.
When I create an EmailTemplate with only Default Content, then accessing a template using a language may fail. This PR fixes that problem.

Additionally I updated some German translations.


selwin:
Hello, trying to use an EmailTemplate with non existing language will trigger an error. This is by design and not a bug you introduced :). Please submit an update for German translations in a separate PR. Thanks!

jrief:
No, this is a bug, please merge it. If you create a template, the default content is with ``language=None``. This should be the fallback in case one uses another language. Otherwise the admin always would have to add at least two templates and the default one would not be be good for anything.

PR #111, opened 2015-07-27
Set Email.template even when render_on_delivery=False

### LINKED ISSUE #110: Email.template not set when using render_on_delivery=False
When calling send() with `render_on_delivery=False`, the resulting Email object's `template` field is not being set.
Is this on purpose?

It could be useful for tracking how many emails with this template were sent, etc...

yprez:
Useful for tracking past emails

Fixes #110

### DIFF
--- a/post_office/mail.py
+++ b/post_office/mail.py
@@ -82,7 +82,7 @@ def create(sender, recipients=None, cc=None, bcc=None, subject='', message='',
             html_message=html_message,
             scheduled_time=scheduled_time,
             headers=headers, priority=priority, status=status,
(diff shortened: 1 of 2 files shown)


yprez:
Tests are currently failing because of:

```
self.assertEqual(email.template, None)
```

So back to the question I asked in #110, is this by design? Is it still necessary?

yprez:
This currently breaks template rendering when render_on_delivery=False, I'll close this for now.

@selwin , I can probably fix this, if there is a chance of it getting merged. What do you think?

PR #113, opened 2015-08-04
Fix missing app_label deprecation warnings (Django 1.9)

yprez:
Fixes the warning:

> /lib/python3.4/site-packages/post_office/models.py:220: RemovedInDjango19Warning: Model class post_office.models.Attachment doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.


selwin:
Thanks!

utapyngo:
When will this be released to pypi?

selwin:
Will cut a new release when https://github.com/ui/django-post_office/pull/118 is merged in.

PR #118, opened 2015-09-16
Bugfix for SMTPServerDisconnected

zwack:
Occasionally we had failed mails with the exception `SMTPServerDisconnected` and message `"please run connect() first"`. We noticed that this happened when calling `post_office.mail.send` with `priority='now'` only and not when using the post_office management command.

After some digging and trying to replicate the problem in the shell (which worked fine using `'now'` at first) I found out that the session is not closed after an email has been sent. So I guess this depends on the email provider you use, but for us after some time the remote server dropped the connection. When trying to send through this connection than, naturally the `SMTPServerDisconnected` exception occurs.

I could only find the django `send()` call once and added a `connection.close()` when it was successful. So far this fixed the issue.

### DIFF
--- a/post_office/mail.py
+++ b/post_office/mail.py
@@ -240,7 +240,8 @@ def _send_bulk(emails, uses_multiprocessing=True, log_level=None):
 
     try:
         for email in emails:
-            status = email.dispatch(log_level=log_level)
+            status = email.dispatch(log_level=log_level,
+                                    disconnect_after_delivery=False)
             if status == STATUS.sent:
                 sent_count += 1
                 logger.debug('Successfully sent email #%d' % email.id)
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -114,8 +114,9 @@ class Email(models.Model):
         if log_level is None:
             log_level = get_log_level()
 
-        connection = connections[self.backend_alias or 'default']
+        connection = None
         try:
+            connection = connections[self.backend_alias or 'default']
             self.email_message(connection=connection).send()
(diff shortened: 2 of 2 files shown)


r4ts0n:
I encounter the same problem when sending via Gmail using Celery.

Hope this will soon be merged and solves the issue.

sanderheling:
We also had the same problem using Mandrill. They only support 10 simultaneous connections.

Maybe it is better to remove the `else` and put the close below the try/catch, because you also want to close the connection when sending the mail has failed.

zwack:
Thought about that but the connection is retrieved in the `try` block. If that fails `connection` will not be available.
One way would be to move retrieving the connection outside of the `try` block and then catch an invalid alias seperately.

selwin:
Hello, this PR will greatly slow down batch email sending because it will need to re-establish an SMTP connection for each email that it needs to send. See https://docs.djangoproject.com/en/1.8/topics/email/#sending-multiple-emails

The easiest way we can work around this is to add a `disconnect_after_delivery` argument to `email.dispatch()`. The `_send_bulk` function would then need to call `email.dispatch` with `disconnect_after_delivery=False` to prevent it from closing the connection after each delivery.

selwin:
@zwack mind updating the PR? :)

zwack:
I see. Just saw `send_many` in the code and thought that will take care of bulk mails. Anyway PR is updated as suggested.

selwin:
Thanks for the PR!

PR #123, opened 2015-11-05
add missing migrations. fixes #122

### LINKED ISSUE #122: Your models have changes that are not yet reflected in a migration, and so won't be applied.
If i launch a makemigrations i see a new post_office migration:
0003_auto_20151105_2015.py

I think you forgot adding the migration in the latest commit

ScottBarkman:
Feel free to rename the migration file - but figured i'd throw this into a pull request for you

### DIFF
--- a/post_office/migrations/0003_auto_20151105_1453.py
+++ b/post_office/migrations/0003_auto_20151105_1453.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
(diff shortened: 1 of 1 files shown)


hsoft:
I've had a similar problem, but this is not a problem of missing migrations. The migration in this PR is specific to your own setup, @ScottBarkman, so this can't be merged as is. We need to find a way to "silence" migration generation on changes in the "choice" attribute of `EmailTemplate.language`.

I don't know where your second AlterField comes from though.

selwin:
I released a new version on PyPI yesterday so please check that one out and see if it works for you.

Issue #125, opened 2015-12-02
Django 1.9 compatibility issues


yprez:
So far I found the following issues:

1. The imports in `post_office/__init__.py` are causing: `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.` in validation, before the tests are running.

2. The structure of `django.core.mail.backends.base.BaseBackend` changed somehow and the import is broken, easy fix.

3. `django.utils.log.dictConfig` was removed. Can maybe replaced with `logging.config.dictConfig` (Python 2.6 support must be dropped), or doing something like: https://github.com/nvie/rq/commit/77fb32791d6fa219b4b08469e252cedbd08f964b. Since Django 1.9 isn't supposed to work with Python 2.6, the best would be to fallback from one to another.

4. Django template error messages were changed causing failed assertions in test_invalid_syntax() - easy fix.

I can fix 2 - 4, but no idea what to do with 1.

PR #126, opened 2015-12-02
Django1.9 support

yprez:
The only "breaking" change here is removing `from .utils import send_mail` in `__init__.py`. But it's not documented anywhere, not sure if it was used before.

Ref #125


selwin:
Thanks for the PR!

yprez:
Awesome! Thanks for the quick review!

Issue #135, opened 2016-01-29
Suggestion: __str__ methods and explicit __unicode__


PukeCloud:
Hello.

It would be nice if models had `__str__` methods since python 3 django uses those methods instead of `__unicode__` methods. Thus all objects in admin interface are labeled as `<Model> object`.

Also, it would be great if `EmailTemplates` that have `default_template` had different return of `__unicode__` method (i.e. `"%s: "%s" % (self.default_template, self.language)`). Otherwise in admin interface you have to pick from several templates that have exactly the same label.

yprez:
True, str() doesn't seem to work in Python 3.

e.g.

Python 3:
```
>>> EmailTemplate(name='123')
<EmailTemplate: EmailTemplate object>
```

And in Python2:
```
>>> EmailTemplate(name='test')
<EmailTemplate: test>
```

PR #136, opened 2016-01-29
Fix and improve model str() method

yprez:
Ref #135

* str(EmailTemplate) was broken for Python 3. Replaced `__unicode__` with `__str__` and a `@python_2_unicode_compatible` decorator.
* Add language to `EmailTemplate.__str__()` method.
* Test repr() of models.


selwin:
Thanks :)

PR #137, opened 2016-01-30
Test on Python 3.5

yprez:
* Test on Python 3.5
* Always install the latest bugfix version of Django
* Fix some deprecation warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
 
 env:
   - DJANGO=1.4.20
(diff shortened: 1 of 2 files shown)


selwin:
Thanks for th PR!

yprez:
@selwin btw, what do you think about dropping support for older versions of Django and Python 2.6?

There are Django 1.10 warnings (that will become errors on 1.10) that I think would be a pain to resolve otherwise...

selwin:
I'm ok with dropping older versions of Django and Python. I stopped supporting Django 1.4 because of the url dispatching mechanism changes in my other projects. I don't mind doing something similar here to make future maintenance easier.

PR #138, opened 2016-02-04
Allow longer subjects in Email model

### LINKED ISSUE #99: Subject field is too short
Django AdminEmailHandler returns up to 989 symbols in subject.post_office has only 255 symbols limit, so logging of long string raises exception.

yprez:
According to email RFC

Fixes #99 

\* I left the EmalTemplate subject field as is on purpose, because subjects longer than 80 chars are generally considered a bad practice (so 255 chars should be more than enough).


selwin:
Thanks!

PR #139, opened 2016-02-06
Email.dispatch() - catch exceptions more carefully

yprez:
* Don't use empty "except:" block
* Catch exception explicitly, instead of relying on `sys.exc_info()`

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-import sys
 from uuid import uuid4
 
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


selwin:
Does this give detailed traceback though? From my understanding, it only outputs the exception name.

yprez:
I was trying to replicate the previous behavior here...

`sys.exc_info()` is returing a tuple of `(type, value, traceback)` (`value` is the exception object not the message).
In the base version the exception goes into `message`, it's type into `exception`, and the traceback is ignored.

This is quite old, so if I'm not missing anything the traceback never went into the db, only the exception name and message.

Do you think it's better to store the entire traceback in `Log.message`?
(it doesn't make this much harder, I just didn't want to mess up the backwards compatibility)

selwin:
Got it, thanks for cleaning this up :)

PR #140, opened 2016-02-09
replaced == with startswith

bashu:


selwin:
Is there any reason why this change is needed?

bashu:
For example:

https://github.com/edoburu/fluentcms-emailtemplates/blob/master/fluentcms_emailtemplates/views.py#L44-L47

I have two options either patch other apps or postoffice, which one is the best?

selwin:
No worries, I'll pull this in. Just wanted to know the reasoning behind this PR :)

Issue #143, opened 2016-02-16
Suggestion: support for standard django templates


litchfield:
Surely we want to use django templates, rather than dicking about with ugly html strings in code/admin?

yprez:
There are a couple of other apps that use that approach.
I used a few several years ago, and remember it being a total mess. Maybe there are better things that appeared since then (I really don't remember what I used before post office, it was too long ago).

Using Django templates usually creates it's own mess. You'll need 3 template files for every email (subject, text content, html content) and a specific directory structure. Handling errors is also complicated.

What I did in my last project, was creating a management command that loads django templates from files and creates/updates post_office EmailTemplate objects based on them.
This way I'm keeping the templates in version control, and load them to the DB when deploying, so all the errors are also caught there.

I can post here the code for the management command if it's useful to anyone.
I doubt it's something that @selwin will want in the project... but maybe it can be improved a bit and packaged into an external dependency.

yprez:
Also, I think that [django-mailviews](https://github.com/disqus/django-mailviews) does what you're looking for (didn't get to use it myself)...

selwin:
My strategy is actually similar to @yprez .

I think we can do is ship a management command that takes a template name as an input and converts it to a post office template.

yprez:
@selwin that sounds great.

I can work on that, I already have a sort of a prototype (and less non application specific code in my project is also nice :))

I'm using a directory structure that looks like:
```
/templates/
├── email_template_name
│   ├── content.html
│   ├── content.txt
│   └── subject.txt
```
I can reuse that, or make something new.

The hardest part will probably be writing the docs for it...

koriaf:
But won't it create 2 places with same data?... If someone edits stored text files it's required to run load command on production environment. And production version of this email already can be modified by some user. And end user, who is staff member, may be confused by 10 KB of complex HTML markup (while just wanted to change person name in email signature).

I did the same but different way: just manually get EmalTemplate from the database, merge it with local stored template file and send as typical provided template. https://gist.github.com/koriaf/9dab35e0bdd11e766447

jucajuca:
This feature would be indeed really nice to have. Some time ago I had to solve it like below.

It would be much, much cleaner if I can ship the templates with the application. 


https://github.com/ui/django-post_office/pull/357

```
from django.apps import AppConfig
from post_office.apps import PostOfficeConfig
from django.db.models.signals import post_migrate


def create_emails(sender=None, **kwargs):
[...]

PR #149, opened 2016-03-04
Load templates command - first iteration for review

yprez:
Management command for loading Django templates into post office EmailTemplate objects, as discussed in #143.

The `load_email_template` command will use Django's template loaders to look for matching email templates. The template loaders are reused, so the email templates can be put in any folder set up for template collection (template dirs / app dirs / etc.).

A a custom template directory structure.

For example:
`manage.py load_template foo` will look for the following templates:

/foo/subject.txt
/foo/content.txt
/foo/content.html

This will create an EmailTemplate named "foo", with the subject, content, html_content populated according to the files.

@selwin if you can review it would be great. The next step will be writing documentation for it.


yprez:
And I need to figure out how to make it work for Django versions older than 1.9

post_office/management/commands/load_email_template.py
```diff
@@ -0,0 +1,32 @@
...
+
+logger = setup_loghandlers()
+
+
+def _get_template(template_name):
+    """Find the template and return the raw template file content"""
+
+    template = loader.get_template(template_name)
+    content = template.template.source
+    content = content.strip()
```
selwin: I don't think we should call `.strip()`, it could be that the user actually wants to have whitespaces before and after the content, this may affect how `content.txt` or `subject.txt` are displayed on the email.

post_office/management/commands/load_email_template.py
```diff
@@ -0,0 +1,32 @@
...
+from ...models import EmailTemplate
+from ...logutils import setup_loghandlers
+
+logger = setup_loghandlers()
+
+
+def _get_template(template_name):
+    """Find the template and return the raw template file content"""
+
+    template = loader.get_template(template_name)
```
selwin: Does this line fail with a clear error message if `loader` can't find the specified template? In my experience, this could be a source of frustration if the user doesn't know where he's supposed to put the files.

post_office/management/commands/load_email_template.py
yprez: I added it because I had an issue of subject.txt having a `\n` in the end, so it fails when post_office tries to send the actual email.

Will `title.replace('\n', '')` be better?

post_office/management/commands/load_email_template.py
yprez: It throws an exception:
```
$ python manage.py load_email_template foo

Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/raven/contrib/django/management/__init__.py", line 41, in new_execute
    return original_func(self, *args, **kwargs)
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/home/yuri/repos/django-post_office/post_office/management/commands/load_email_template.py", line 29, in handle
    'subject': _get_template(template_name + '/subject.txt'),
  File "/home/yuri/repos/django-post_office/post_office/management/commands/load_email_template.py", line 13, in _get_template
    template = loader.get_template(template_name)
  File "/home/yuri/virtualenvs/inbli-core/lib/python3.4/site-packages/django/template/loader.py", line 43, in get_template
    raise TemplateDoesNotExist(template_name, chain=chain)
django.template.exceptions.TemplateDoesNotExist: foo/subject.txt
```
Maybe a try-except with a more descriptive message will be better.

The bigger problem is that I can't get it to work on older Django versions because of changes to templates, so I can get the template file, but not the file path or the unprocessed text of the template...

Maybe better to drop the the template loaders, and instead to add a  `POST_OFFICE_TEMPLATES_PATH` setting and just get the files from there...

yprez:
@selwin I changed the way templates are loaded from files (couldn't make it work with template loaders at all). Can you give it another review? Thanks!

post_office/management/commands/load_email_template.py
```diff
@@ -0,0 +1,45 @@
...
+        subject = _get_template(template_name, 'subject.txt').strip()
+        content = _get_template(template_name, 'content.txt')
+        html_content = _get_template(template_name, 'content.html')
+
+        fields = {
+            'subject': subject,
+            'content': content,
+            'html_content': html_content,
+        }
+        template, created = EmailTemplate.objects.get_or_create(
```
selwin: Using `update_or_create` here would simplify the code a bit.

post_office/management/commands/load_email_template.py
selwin: Since we changed the template loading mechanism to a simple `open` instead of going through Django's loader, the exception no longer looks like this.

I think you can read the source by reading the file specified in `template_instance.origin.name` as suggested in this post: http://stackoverflow.com/questions/22256831/get-the-source-of-a-django-template

I don't mind either source reading mechanism. In your opinion, which kind of exception is better to show? Django's `TemplateDoesNotExist`, Python's `IOError` or should we make our own?

yprez: Maybe catching the exception and just printing a message explaining the required directory structure, setting, etc.?

post_office/management/commands/load_email_template.py
yprez: I removed it in f01eff774 to support older versions of Django (< 1.7)...

post_office/management/commands/load_email_template.py
yprez: The errors now look like this:
```
$ python manage.py load_email_template some_template_name

[Errno 2] No such file or directory: 'email_templates/some_template_name/subject.txt'

Please make sure you defined the correct email template directory structure.
You can change the base directory ("email_templates") by setting POST_OFFICE['TEMPLATE_DIR'] in your settings file
```

what do you think?

selwin:
Sorry, I've been really busy lately. I'll try to go over this again in a few days.

post_office/management/commands/load_email_template.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Load an EmailTemplate from Django template files'
+    args = 'template_name'
+
+    def handle(self, template_name, *args, **options):
+        base_dir = settings.get_template_dir()
+
+        def _get_template(filename):
```
selwin: Minor nitpick, can we rename this to `get_content(filename)`?

selwin:
Overall, this PR looks great. Could you also document this feature in README?

post_office/management/commands/load_email_template.py
yprez: And now I can revert it because support for older Django versions is dropped :)

yprez:
@selwin thanks for the review... I didn't have time to finish this yet, but I'll get to it at some point

PR #152, opened 2016-03-24
Fix POST_OFFICE_CACHE and POST_OFFICE_TEMPLATE_CACHE behavior

koriaf:
before you need to set both variables, first to anything, and second to False, to avoid caching. Now any variable, set to False, will stop cache. It's useful when you are having json-backed cache backend, which just can't save EmailTemplate model without special actions.

To test it manually update test_settings.py to different values of these variables and run tests, and see how and when they fail.

### DIFF
--- a/post_office/test_settings.py
+++ b/post_office/test_settings.py
@@ -7,6 +7,10 @@ DATABASES = {
     },
 }
 
+# Default values: True
+# POST_OFFICE_CACHE = True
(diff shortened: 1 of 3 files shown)


selwin:
Do you mind writing an automated test for this?

koriaf:
You are welcome.

selwin:
Thanks!

PR #153, opened 2016-04-04
Changed choices for language field of EmailTemplate model to be dynamic

desecho:
Changed choices for language field of EmailTemplate model to be dynamic and set the default value to [] 
to avoid making false migrations if LANGUAGES settings is changed. 
Related to https://github.com/ui/django-post_office/pull/123#issuecomment-161747378


selwin:
Thanks!

PR #156, opened 2016-04-27
How to use uwsgi timers instead of cron

jrief:
This alternative to cron was very useful for me. Therefore I'd like to share this knowledge with others.


selwin:
Thanks!

Issue #150, opened 2016-03-11
Django 1.9: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.


utapyngo:
lib/python2.7/site-packages/django/utils/six.py:808:
      RemovedInDjango110Warning: SubfieldBase has been deprecated.
      Use Field.from_db_value instead.

SubfieldBase is used by `CommaSeparatedEmailField`, see
https://github.com/ui/django-post_office/blob/e79dc8f304c80b6982178fa323760d1fc29f436c/post_office/fields.py#L9

selwin:
Do you mind making a PR for this?

utapyngo:
I don't know how to fix this correctly.

PR #161, opened 2016-05-17
SubfieldBase has been deprecated. Use Field.from_db_value instead.

simonphilips:
Should fix issue #150.

`from_db_value` turns a database value into a Python object
`to_python` is called during form.clean()

### DIFF
--- a/post_office/signals.py
+++ b/post_office/signals.py
@@ -8,6 +8,6 @@ from .fields import CommaSeparatedEmailField
 
 @receiver(post_init, sender=Email)
 def on_email_post_init(sender, instance, **kwargs):
-    for field in {"to", "cc", "bcc"}:
+    for field in ("to", "cc", "bcc"):
         value = CommaSeparatedEmailField.parse_value(getattr(instance, field))
         setattr(instance, field, value)
--- a/post_office/signals.py
+++ b/post_office/signals.py
@@ -1,0 +1,13 @@
+
+from django.db.models.signals import post_init
+from django.dispatch import receiver
+
+from .models import Email
+from .fields import CommaSeparatedEmailField
+
+
+@receiver(post_init, sender=Email)
+def on_email_post_init(sender, instance, **kwargs):
+    for field in {"to", "cc", "bcc"}:
+        value = CommaSeparatedEmailField.parse_value(getattr(instance, field))
+        setattr(instance, field, value)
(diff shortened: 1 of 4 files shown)


selwin:
Thanks for the Pr. Could you please check why tests are failing?

simonphilips:
I'm not sure if the `test_models_repr` test is entirely valid. Django does not validate field data when creating Model(field=value). Validation is done in the forms.

For example:

```python
class TestModel(Model):
    number = IntegerField()

test = TestModel(number="test string")
```

Will not result in an error. `test.save()` does raise a ValueError however.

The unit test should probably check the Field methods instead.

As for backwards compatibility, if both a list and a string should always be transformed to a list after creating (but not yet saving) Email, then this processing should probably be done in a post_init signal for Email instead. 

I've updated the patch.

simonphilips:
Everything should work now. After dealing with some exotic Python/Django combinations all tests finally pass.

post_office/signals.py
```diff
@@ -0,0 +1,13 @@
+
+from django.db.models.signals import post_init
+from django.dispatch import receiver
+
+from .models import Email
+from .fields import CommaSeparatedEmailField
+
+
+@receiver(post_init, sender=Email)
+def on_email_post_init(sender, instance, **kwargs):
```
selwin: Can we set the value attribute in `CommaSeparatedEmailField`'s `__init__` method? It seems more intuitive that way.

simonphilips: The problem is that `Email.__init__` does not execute `CommaSeparatedEmailField.__init__`. So the following unit test never runs CommaSeparateEmailField's processing and therefore fails:

```python
 def test_models_repr(self):
        self.assertEqual(repr(Email(to='test@example.com')), "<Email: ['test@example.com']>")
```

The output of repr() is simply `<Email: 'test@example.com'>`. The preprocessing/validation must be done in a form instead and `Email.__init__` simply expects the one and correct `['test@example.com']` list value from its caller.

The only way to make that unit test pass is to either override `Email.__init__` (which Django does not recommend) or to use signals as I did above. Alternatively, maybe it's best to consider the test as broken because it tests something that Django doesn't do. 

Consider the following example:

```python
from django.db.models import Model, IntegerField
class MyModel(Model):
    number = IntegerField()
    def __repr__(self):
        return "MyModel(number=%s)" % repr(self.number)

my_instance = MyModel(number="obviously_not_an_integer")
print(repr(my_instance))
```

Which will simply output `MyModel(number='obviously_not_an_integer')` even though my_instance.number clearly conflicts with the IntegerField for the same reason.

selwin:
Thanks for the PR. I dropped support for Django < 1.8 to make the code simpler. Fixed in https://github.com/ui/django-post_office/pull/167

PR #158, opened 2016-05-08
Migrations, translations, bugs...

fizista:
So...

I cleaned a little code, fixed a few bugs and added translation. I've also added the ability to run tests using TOX.

All tests run for Django from version 1.4 to 1.9, and Python versions 2.6 to 3.5.

1) Removed language choices from the model EmailTemplate.
Fixes a bug missing migration. There was a problem with the configuration parameter LANGUAGES.
2) Added missing South migrations.
3) Updated gitignore.
4) The new translation into Polish
5) Add TOX
Added testing environment similar to Travis. Tests can be run at home.
https://pypi.python.org/pypi/tox
6) FIX: DeprecationWarning: Please use assertEqual instead.

PR #165, opened 2016-06-19
Use Tox for tests with different python and django version

koriaf:
Allows people to run tests old-style, as usual, with current python and django version, or use tox automation project to check against different ones. Especially useful for trying things in python 2 if you use python3 and vise versa.


selwin:
Fixed in https://github.com/ui/django-post_office/pull/158

PR #166, opened 2016-07-01
now really fixes #164

### LINKED ISSUE #164: 'bytes' object has no attribute 'encode'
I've tried many things, but it seems like every time I try to send an EMail with an attachment, it fails with the following error in the Log:

     'bytes' object has no attribute 'encode'

Apparently there was a Django bug, with exactly the same problem (https://code.djangoproject.com/ticket/24623), but it seems to have been fixed:
https://github.com/django/django/commit/c6da621def9bc04a5feacd5652c7f9d84f48ad2c

I'm using both the newest version of Django(1.9.7) and django-post_office(2.0.7) and running on Python Python 3.5.1. This is also happening on two different machines both running Linux.

Here some sample code:

    from post_office import mail
    mail.send(
        'some@email.com',
        'some@email.com',
        subject='My email',
        message='Hi there!',
        attachments={
            'manage.py': 'manage.py',
        },
        priority='now',
    )

Brandl:
Most of the information concerning the bug an possible fixes can be found here:
https://code.djangoproject.com/ticket/26802

As long as the Django project does not come up with a simple solution, I think this fix is the best way mitigating the problem. I will watch the development closely and may provide a cleaner solution if there is a fix.

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -98,7 +98,7 @@ class Email(models.Model):
                 connection=connection, headers=self.headers)
 
         for attachment in self.attachments.all():
-            msg.attach(attachment.name, attachment.file.read())
+            msg.attach_file(attachment.file.path)
 
         return msg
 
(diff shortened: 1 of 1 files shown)


jrief:
closing for inactivity

PR #175, opened 2016-10-08
adding success/failure logging

dinopetrone:
apologize for previous pull request without proper testing.  All this does is log out if there is an error when sending out an email.  I have also adjusted models.py to be pep8 compliant.

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -27,6 +27,7 @@ PRIORITY = namedtuple('PRIORITY', 'low medium high now')._make(range(4))
 STATUS = namedtuple('STATUS', 'sent failed queued')._make(range(3))
 
 
+logger = logging.getLogger(__name__)
 
 @python_2_unicode_compatible
 class Email(models.Model):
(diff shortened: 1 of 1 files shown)


dinopetrone:
ping

ktosiek:
I want this too, and I'd even say sending an email is an INFO, not DEBUG :-)

But mixing this with PEP8 changes seems like an overkill.

post_office/models.py
```diff
@@ -138,6 +143,11 @@ def dispatch(self, log_level=None, disconnect_after_delivery=True):
         self.status = status
         self.save()
 
+        if status == STATUS.failed:
+            logger.error("email send fail: %s", self)
```
selwin: The log level for this should be "info" instead of error because this is not an error in this library. Also, please change the message to "Delivery failed: "

post_office/models.py
```diff
@@ -138,6 +143,11 @@ def dispatch(self, log_level=None, disconnect_after_delivery=True):
...
+        else:
+            logger.debug("mail status: %s \nfor email: %s", status, self)
```
selwin: Please change the logging level to "info" and copy to "Delivery successful: "

post_office/models.py
ktosiek: Will those failed emails be retried automatically? If not, then I think it's at least a "warning" - as it means something is going wrong, even if it's not an "error" as it doesn't need immediate attention.

selwin:
Sure, please change the level to warning. Thanks!

PR #181, opened 2017-02-07
Dropping support for Django < 1.8

fendyh:
This commit is to update the project dependencies and read file as it
no longer supports Django < 1.8

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,11 +41,7 @@ Installation
         "post_office",
     )
 
-* Run ``syncdb`` for Django == 1.8::
-
-    python manage.py syncdb
-    
-* Run ``makemigrations`` and ``migrate`` for Django > 1.8::
+* Run ``makemigrations`` and ``migrate``::
 
     python manage.py makemigrations
     python manage.py migrate
--- a/README.rst
+++ b/README.rst
@@ -41,9 +41,8 @@ Installation
         "post_office",
     )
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -41,7 +41,7 @@ Installation
         "post_office",
     )
 
-* Run ``syncdb`` for Django <= 1.8::
+* Run ``syncdb`` for Django == 1.8::
```
yprez: Actually, I think `python manage.py migrate` exists and `syncdb` is deprecated in 1.7+,  so this can be removed...

fendyh: Yes, I agree with you, I just noticed this because I've been using Django 1.7 in all my projects and have not started a new one with Django 1.8 yet. Thanks for the correction @yprez , I will now commit a new one which removes ``syncdb``

README.rst
```diff
@@ -41,11 +41,7 @@ Installation
         "post_office",
     )
 
-* Run ``syncdb`` for Django <= 1.8::
-
-    python manage.py syncdb
-    
-* Run ``makemigrations`` and ``migrate`` for Django > 1.8::
+* Run ``makemigrations`` and ``migrate``::
```
selwin: `makemigrations` shouldn't be needed.

fendyh: done in new commit, removed `makemigrations` as it is unnecessary

selwin:
@fendyh can you also remove `south_migrations`?

PR #187, opened 2017-04-20
fix EmailTemplate so that it can be created more than once

sac7e:
the current value of `unique together` under `EmailTemplate` makes it impossible to have two more templates. given name variable is required in `EmailTemplate`, it is fine to change the `unique_together` from `(language, default_template)` to `(name, language, default_template)`


selwin:
You'll also have to create a migration for this.

PR #191, opened 2017-06-09
Support for explicit attachment mimetypes

### LINKED ISSUE #155: Should be able to specify attachments' mime types
I'm using Sparkpost's email backend (https://github.com/SparkPost/python-sparkpost) along with post office. The problem is that Sparkpost's api seems very strict and requires each attachment to have mime type specified. It seems to me that there's no way to specify attachments' mime types in post office's API. Right now, I'm not able to use post office and Sparkpost out of box at this moment.

clickonchris:
Fixes #155 

This change allows the caller to pass a mime type with each attachment, which will in turn get passed to Django and then to the EmailBackend.

I made this change because our email provider (Postmark) requires an explicit mime type or content-type to be sent with the attachment.  Django supports it (core/mail/message.py::attach)  but post_office did not.

Django Ref: https://docs.djangoproject.com/en/1.11/topics/email/#emailmessage-objects


post_office/models.py
```diff
@@ -242,6 +242,7 @@ class Attachment(models.Model):
     name = models.CharField(_('Name'),max_length=255, help_text=_("The original filename"))
     emails = models.ManyToManyField(Email, related_name='attachments',
                                     verbose_name=_('Email addresses'))
+    mimetype = models.CharField(max_length=255, null=True)
```
selwin: Could you please change this to `models.CharField(max_length=255, blank=True, default='')`?

clickonchris: My thinking here is that django.core.messages.py::EmailMessage.attach is expecting None by default, so passing empty string if no value is explicitly set (instead of None) would do more harm than good.

BTW it looks like django 1.11 has some support for automatic mimetype detection if mimetype is None.

1.8 (1.9, 1.10 have no significant changes)
https://github.com/django/django/blob/1.8.18/django/core/mail/message.py#L305

1.11
https://github.com/django/django/blob/1.11.2/django/core/mail/message.py#L350

selwin: I think the best practice is to avoid having `None` in `CharField` unless it's really needed to avoid having two representation of empty values, `None` and empty string.

On the Django side, we can simply do `msg.attach(attachment.name, attachment.file.read(), mimetype=attachment.mimetype or None)` so `msg.attach` will receive `None` if `mimetype` is empty.

clickonchris: a good solution 👍

post_office/models.py
```diff
@@ -242,6 +242,7 @@ class Attachment(models.Model):
     name = models.CharField(_('Name'),max_length=255, help_text=_("The original filename"))
     emails = models.ManyToManyField(Email, related_name='attachments',
                                     verbose_name=_('Email addresses'))
+    mimetype = models.CharField(max_length=255, null=True, default='')
```
selwin: Sorry for not being clear, we don't want the DB column to be nullable. Could you please change this to `models.CharField(max_length=255, blank=True, default='')`?

clickonchris: no problem at all.  It does make more sense this way.

PR #219, opened 2018-01-08
Switch to a better implementation of JSONField

peterdemin:
There are two similar implementations of JSONField.
Django Post Office uses the first one.

1. Maintainer: Dan Koch
   Source: https://github.com/dmkoch/django-jsonfield
   Current Version: 2.0.2
   PyPI: https://pypi.python.org/pypi/jsonfield

2. Maintainer: Matthew Schinckel
   Source: https://bitbucket.org/schinckel/django-jsonfield
   Current version: 1.0.1
   PyPI: https://pypi.python.org/pypi/django-jsonfield

Both packages are installed into `jsonfield` directory and have the
same API.
However JSONField by Dan Koch doesn't work correctly with
`django.utils.model_to_dict` - it returns dumped JSON string
representation, instead of object.

Here is the code responsible for this [mis]behavior:

https://github.com/dmkoch/django-jsonfield/blob/master/jsonfield/fields.py#L110

```
    def value_from_object(self, obj, dump=True):
        value = super(JSONFieldBase, self).value_from_object(obj)
        if self.null and value is None:
            return None
        return self.dumps_for_display(value) if dump else value
```

It dumps value by default, while model_to_dict expects it
to return value unchanged.

JSONField by Matthew Schinckel doesn't override `value_from_object`
and works as Django expects.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
     zip_safe=False,
     include_package_data=True,
     package_data={'': ['README.rst']},
-    install_requires=['django>=1.8', 'jsonfield'],
+    install_requires=['django>=1.8', 'django-jsonfield'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Environment :: Web Environment',
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ setenv =
 
 [testenv]
 deps =
-  jsonfield
+  django-jsonfield
   django18: Django >= 1.8, < 1.9
   django19: Django >= 1.9, < 1.10
   django110: Django >= 1.10, < 1.11


selwin:
I'm quite hesitant to pull in this change because this may break existing installations in unexpected ways. Have you tried opening an issue and see if the author can get this issue fixed?

peterdemin:
Hi Selwin, no, I didn't. I've read git logs for this method and saw, that it had this strange behavior from the beginning. But sure, I'll try to reach out to Dan Koch.
I understand your concern. But think about it - Django Post Office's tests pass just fine with alternative library and it has exactly the same interface. What if you try to make a (beta) release and see if it breaks anything?

peterdemin:
JFYI Here is my PR to currently used django-jsonfield: https://github.com/dmkoch/django-jsonfield/pull/213
It turned out to be easier to fix than I first thought.

phoebebright:
Having both django's own json field (postgres only) and jsonfield can cause strange behaviour with data being re-encoded as string.  See https://bitbucket.org/schinckel/django-jsonfield/issues/57/cannot-use-in-the-same-project-as-djangos.  Will do a fork to use django's Jsonfield for my own project but that is not a good solution for anyone using mysql.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/285

jrief:
Just a short reminder: In Django-3.1 there will be a database independent JSONField. This fortunately will fix the current mess with the different JSON field implementations.

Mogost:
> Just a short reminder: In Django-3.1 there will be a database independent JSONField. This fortunately will fix the current mess with the different JSON field implementations.

I think that JsonField will be a problem, at least until the end of support for Django 2.2 (April 2022)

PR #222, opened 2018-01-29
Add suspend/resume capabilities to background sending of emails

SeiryuZ:


post_office/models.py
```diff
@@ -80,6 +83,28 @@ def __init__(self, *args, **kwargs):
     def __str__(self):
         return u'%s' % self.to
 
+    @classmethod
+    def suspend(self):
+        """ Create a lockfile that suspend background sending of emails.
```
selwin:
```
Creates a lockfile that suspends background sending of emails.
```

post_office/models.py
```diff
@@ -80,6 +83,28 @@ def __init__(self, *args, **kwargs):
...
+            User is responsible for clearing up the suspend lockfile.
+        """
+        open(SUSPEND_FILEPATH, 'a').close()
+
+    @classmethod
+    def is_suspended(self):
+        """ Check if the suspend lockfile exist or not
```
selwin: `Check whether email delivery is currently suspended`

post_office/management/commands/suspend_mail.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management.base import BaseCommand
+
+from ...models import Email
+
+
+class Command(BaseCommand):
```
selwin: `suspend_email_delivery`

post_office/management/commands/resume_mail.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management.base import BaseCommand
+
+from ...models import Email
+
+
+class Command(BaseCommand):
+    help = 'Resume queued email sending.'
+
+    def handle(self, **options):
```
selwin: `resume_email_delivery`

selwin:
Instead of suspending/resuming email delivery, I have another project which does something like this: ```python
# This means mail will only be delivered during 8am to 8pm
mail.send(foo, bar, delivery_window(time(08, 00), time(20, 00))
```

Let's do this instead.

jrief:
@selwin a good library to handle datetime ranges is here: https://pypi.org/project/DateTimeRange/

Issue #223, opened 2018-02-02
post_save of Log not working 


subramanyamVemu:
when we the email with send_queued_mail in a cron

In 2.0.8 we were receiving post_signal of Log

However in 3.0.3/4 we are not receiving the post_save signal of Log

subramanyamVemu:
Any work around will help since I have got it on my production box !!

subramanyamVemu:
Found the issue
https://github.com/ui/django-post_office/blob/62aa4aa516e9ae87d700fb4dee43c473aceb7434/post_office/mail.py#L287

and in

https://github.com/ui/django-post_office/blob/62aa4aa516e9ae87d700fb4dee43c473aceb7434/post_office/mail.py#L297

we are doing a bulk_create which doesn't raise the signals
Please review 

Can I go ahead and send a merge request

blag:
While this issue is seven years old at this point, this project still uses `Log.objects.bulk_create(logs)`:
https://github.com/ui/django-post_office/blob/d8f65bed493bb1d8bb02f07677eb9c51c8b86cbf/post_office/mail.py#L428
https://github.com/ui/django-post_office/blob/d8f65bed493bb1d8bb02f07677eb9c51c8b86cbf/post_office/mail.py#L435

However, creating logs in bulk is far more efficient than creating them one-by-one.

I think this issue can be closed.

PR #229, opened 2018-06-28
Issue #223 -post_save signal of log not triggered

subramanyamVemu:


selwin:
Moving from bulk creations to individual `save()` calls will significantly reduce performance on large datasets so I won't accept this PR.

If you have a use case that relies on listening to signals, please open an issue and we'll see if we can find alternative solutions.

PR #232, opened 2018-07-17
Attachments subdirectory

### LINKED ISSUE #230: Attachments shouldn't be put into one huge directory
On some system there is a limit on number of files in a directory. It might be wise to put the attachments in some kind of sub-directory structure compared to the current approach of dumping everything down to `post_office_attachments` directory.

`post_office_attachments/{year}/{month}/{date}/{attachment_file}` (Y-M-D on attachment creation) might be good alternative ?

SeiryuZ:
Fix #230


SeiryuZ:
Whoops wrong branch

PR #240, opened 2018-08-10
Catch ValueError when trying to read lockfile

bsoist:
We're not sure why exactly, but at thinkeasy.net we have occasionally seen an error trying to `int` the contents of the lock file because it's empty. I thought this might be a helpful addition.


selwin:
Thanks!

PR #245, opened 2018-10-01
add highlighting for email templates

will-emmerson:
A small modification to the admin change_form to add syntax highlighting to subject, content and html_content:

screen shot 2018-10-01 at 16 37 00

Not sure if this is the best way to do it, perhaps Form.Media would be better?


selwin:
Thanks!

Issue #246, opened 2018-10-04
Unbound local variable `recipient` error in named empty email validation


Praseetha-KR:
In the case of named empty email (eg: `Alice <>`), I'm getting the following error: 

```bash
File "/usr/local/lib/python3.6/dist-packages/post_office/validators.py", line 25, in validate_email_with_name
   validate_email(recipient)
UnboundLocalError: local variable 'recipient' referenced before assignment
```

PR #247, opened 2018-10-04
Unbound var recipient in validators for named empty email case

Praseetha-KR:
- Fix for #246
- Also added test case for named empty email


selwin:
Thanks!

PR #249, opened 2018-11-16
Update CommaSeparatedEmailWidget method override name

reinbach:
The initial name of the method override (`_format_value`) appears to have been incorrectly named. It should be `format_value`

See https://github.com/django/django/blob/master/django/forms/widgets.py#L205


selwin:
Thanks!

PR #251, opened 2018-11-30
Attachments allowing mime base

jrief:
This PR keeps track on the headers in attachments. Using this, one can create HTML Emails with inlined images and django-post_office to postpone delivery.

It is based on https://github.com/dldevinc/django-post_office/tree/patched which seems to be the canonical branch for the latest 3.1.0 version.


jrief:
shall this pull request die in silence?

selwin:
Sorry, I just saw this PR. This PR looks good! Could you please add a brief section in README showing how to send an email with inlined images?

jrief:
In pull request #260 I added a whole section. Basically this feature works by using a special template engine, which keeps track of inlined images. So one can render HTML emails just as he would do with Web-HTML. I added a long section to the README, which explains it in detail.

I needed this feature in django-SHOP's email templates, [example](https://github.com/awesto/django-shop/blob/releases/1.0.x/shop/templates/shop/email/order-detail.html), to render a list of products together with their images. Previously one had to add absolute URLs and hope that the email client would load them from the server. Now, all images are inlined and send together with the mail body.

selwin:
Thanks!

Issue #201, opened 2017-08-08
Use sys.platform to select the type of lock


ktosiek:
On Python 3.2+ `os.symlink` exists even on Windows, which causes permission errors - creating symlinks is a privileged operation there. It would be better to use `sys.platform == 'win32'` to select the right lock implementation.

BTW there doesn't seem to be a Windows CI for the project - would you like one? I've used AppVeyor before and it's pretty nice. They do have a free-for-OSS tier.

PR #253, opened 2018-12-19
Added a check to see if platform is Windows

dhwanils95:
Your 'lockfile.py' has a condition that if the os library of python has attribute 'symlink' then you are trying to create a symlink and clearly in the other condition you have mentioned, it is not supported in Windows. But the OS library in python on Windows does have an attribute 'symlink' but when I try to connect it does not work, as expected. So I have added an additional check to see if the platform is Windows or not.

### DIFF
--- a/post_office/lockfile.py
+++ b/post_office/lockfile.py
@@ -20,6 +20,7 @@
 
 import os
 import time
+import platform
 
 
 class FileLocked(Exception):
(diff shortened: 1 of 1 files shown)


ktosiek:
This should solve https://github.com/ui/django-post_office/issues/201

dhwanils95:
> This should solve #201

Yes, it did solve the issue for me.

selwin:
Thanks!

Issue #252, opened 2018-11-30
Keep published version on PyPI in sync with repository


jrief:
Version 3.1.0 published on PyPI is not available through this repository, although its project link backpoints onto this repository. It seems as if the latest branch is located here: https://github.com/dldevinc/django-post_office/tree/patched

Apparently this happened in earlier version too, see #151 and #131

jrief:
Another reminder to this issue.

WoLpH:
This confused me as well. After using the readme to implement the code I got the error: `No module named 'post_office.template'`

Perhaps it would be a good solution to have a development branch where development happens and a master branch which is always in sync with PyPI?

At the very least, a new release would be appreciated a lot :)

Issue #151, opened 2016-03-22
Version in git is behind pypi?


yprez:
The latest version on pypi is 2.0.6, but it's currently set to 2.0.5 in setup.py.

The github versions of CHANGELOG.md and README.rst also seem to be behind...

selwin:
Sorry about that, I made a release on my computer but forgot to push it to Github. It's fixed now.

yprez:
I think the same is happening with 2.0.7

Issue #131, opened 2015-12-22
Latest version on PyPI is 2.0.4


jrief:
but looking at the repository, the latest ist 2.0.2.
Did you forget to push?

selwin:
I messed up the version numbering while working across computers. I just pushed out 2.0.5. Sorry for the confusion.

yprez:
@selwin I think there's still something wrong, since the github version is `2.0.2` but it's `2.0.5` on pypi...

PR #260, opened 2019-03-24
Inlined images template engine

jrief:
This is an extension of pull request #251. It allows users to create emails using inlined images.
There is a special template engine, which shall be used in combination with a special templatetag. It keeps track on the attached images and adds them as inline-attachments.


jrief:
Some unit tests fail because either Python is too old (<3.5), or Django is too old (<1.10).
I would either have to backport some methods to Django-1.8/1.9 and Python-3.4 or drop these versions. Since it is very unlikely that anyone requires these new features, but still wants to stick with a deprecated version of Python or Django, I would opt for the latter. What do you think?

selwin:
I'm ok with dropping support for Django < 1.11. Would it be troublesome to support Python 3.4 (it's supported by Django 1.11)?

jrief:
> I'm ok with dropping support for Django < 1.11. 

OK, then I will change this in this pull request.

> Would it be troublesome to support Python 3.4 (it's supported by Django 1.11)?

No, it's very simple to backport this.

Give me some time to change it.

jrief:
OK, dropped support for Django<1.10 and Python-3.3.

post_office/template/__init__.py
```diff
@@ -0,0 +1,15 @@
+from django.template.loader import get_template, select_template
+
+
+def render_to_string(template_name, context=None, request=None, using=None):
```
selwin: Is this function only used in tests? If so, please put this in `tests/utils.py`

post_office/templatetags/html_email.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
selwin: Do you mind renaming this file to `post_office.py`? This way post office tags can be loaded with `{% load post_office %}`

post_office/models.py
```diff
@@ -98,30 +96,40 @@ def prepare_email_message(self):
...
-        subject = smart_text(self.subject)
-
         if self.template is not None:
-            _context = Context(self.context)
-            subject = Template(self.template.subject).render(_context)
-            message = Template(self.template.content).render(_context)
-            html_message = Template(self.template.html_content).render(_context)
+            engine = get_template_engine()
+            subject = engine.from_string(self.template.subject).render(self.context)
+            plain_message = engine.from_string(self.template.content).render(self.context)
```
selwin: Minor nitpick, could you please rename this to `plaintext_message`?

post_office/template/backends/html_email.py
```diff
@@ -0,0 +1,54 @@
...
+        return Template(self.engine.from_string(template_code), self)
+
+    def get_template(self, template_name):
+        try:
+            template = self.engine.get_template(template_name)
+            return Template(template, self)
+        except TemplateDoesNotExist as exc:
+            reraise(exc, self)
+
+    def get_templatetag_libraries(self, custom_libraries):
```
selwin: Do we need to override this?

post_office/templatetags/html_email.py
```diff
@@ -0,0 +1,40 @@
...
+from django.conf import settings
+from django.contrib.staticfiles import finders
+from django.core.files import File
+from django.core.files.images import ImageFile
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def image_src(context, file):
```
selwin: Can we please rename this to `inlined_image`?  I also wonder if we can turn this into a template filter instead. So it can be used as such ``

post_office/template/backends/html_email.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import unicode_literals
```
selwin: Can we rename this file to `template/backends/__init__.py` and the engine to `PostOffice`? This way we can refer to this engine as `'post_office.template.backends.PostOffice`

post_office/template/__init__.py
jrief: This function is the counterpart of `django.template.engine.render_to_string` for our special template engine. It is intended to be used by those who want to render the final message themselves and later compose their message.
Therefore I would keep it.

post_office/templatetags/html_email.py
jrief: This is the counterpart of Django's templatetag `static`, therefore I would stick with that convention, rather than using a filter. By using `image_src` I want to emphasize that it is a kind of source object. I also used this naming convention [here](https://github.com/jrief/django-sass-processor/blob/0503663de4ff50e2858bb5e052f0fa96aa1f9003/sass_processor/templatetags/sass_tags.py#L48). 
But I've no objections against renaming it to `inline_image`.

post_office/template/backends/html_email.py
jrief: I can rename it to `template/backends/post_office.py` if you want, but not to `__init__.py`. Reason is that this name is used in the `settings.py` for `POST_OFFICE['TEMPLATE_ENGINE']` and therefore we need a meaningful name here.
This btw. was the main reason, to use `html_email` rather than `post_office` as the template's engine name, because I wanted to emphasize that it is intended for HTML email rendering.

jrief:
Please answer me on the debatable review annotations, so that I can fix them as soon as possible.

post_office/template/__init__.py
selwin: Ok.

post_office/template/backends/html_email.py
selwin: How about moving this to `/post_office/template/backends.py` then? Name the engine `HTMLEmailEngine` or `HTMLEmail` so it can be specified as `'post_office.template.backends.HTMLEmailEngine`.

The built in Jinja2 engine can be imported as `django.template.backends.jinja2.Jinja2`. I think this is a better naming convention than `DjangoTemplates` (it's not a template).

post_office/templatetags/html_email.py
selwin: Ok, let's use `inline_image` template tag then :)

post_office/template/backends/html_email.py
jrief: The problem are these functions: `django.template.loader.get_template`, `django.template.loader.select_template` and  `django.template.loader.render_to_string`. They all take a parameter `using`, which in our case would be `get_template('/path/to/message.html', using='html_email')`. Here `html_email` is the name of the module in `backends`. If we move it to `backends.py`, one would have to call `get_template('/path/to/message.html', using='backends')` – not a very good name.

As said, I have no problem using `post_office` here.

jrief: I just renamed the template engine to `post_office.PostOfficeTemplates`. I would really linke to keep "Templates" inside the name, because that is what Django does with its own template engine: `django.template.backends.django.DjangoTemplates`.

post_office/template/backends/post_office.py
```diff
@@ -20,7 +20,7 @@ def attach_related(self, email_message):
             email_message.attach(attachment)
```
selwin: Can we rename this file from `post_office/template/backends/post_office.py` to `post_office/template/backends.py`?

jrief:
Can I expect any time soon a new release? I'm asking because some of my apps depend on django-post_office and I wanted to know, if I shall handle the project's requirements through PyPI or GitHub.

selwin:
Hi Jacob,

I’m currently out of town. I’ll try to cut a release sometime next week. Is that ok?

Thanks for putting so much effort into this release :)

Best,
Selwin

jrief:
@selwin if you want, I can take over the job of releasing new versions.

I already do this for other Django apps maintained by other people, for instance [easy-thumbnails](https://pypi.org/project/easy-thumbnails/) and [django-treebeard](https://pypi.org/project/django-treebeard/).

I then would also ensure that the Changelog is always up to date, and that version numbers are tagged correctly to the code in the repository. In the past there unfortunately have been various mismatches: https://github.com/ui/django-post_office/issues/252 , https://github.com/ui/django-post_office/issues/151 and https://github.com/ui/django-post_office/issues/131 .

selwin:
@jrief sorry for the delay. Would be glad to have your help in coordinating releases. How do other projects usually do this? Do they just invite you as a collaborator in their respective repos?

jrief:
Yes, on GitHub you can invite someone as collaborator.
On PyPI you can add me as Maintainer.

Be assured that I will never merge new features without your review.
In projects I do not own, I usually only merge bug-fixes.

selwin:
@jrief I have added you on both Github and PyPI. Please check your inbox :).

jrief:
@selwin Thanks, I've seen and accepted.
Shall I now bundle the next minor release, ie. 3.2?

PR #265, opened 2019-05-22
Prevent the need for updating VERSION on two different places

jrief:
In my projects, I prefer to keep only one location which has to be updated on version jumps. Currently it is in 'post_office/__init__.py' and 'post_office/setup.py'.

However, this requires to remove the `from .backends import EmailBackend` from the `__init__.py` file. 
@selwin please tell my if this is OK for you.

### DIFF
--- a/post_office/__init__.py
+++ b/post_office/__init__.py
@@ -1,5 +1,3 @@
 VERSION = (3, 2, 1)
 
-from .backends import EmailBackend
-
 default_app_config = 'post_office.apps.PostOfficeConfig'
--- a/post_office/settings.py
+++ b/post_office/settings.py
@@ -42,7 +42,7 @@ def get_available_backends():
 
     # If EMAIL_BACKEND is set to use PostOfficeBackend
     # and POST_OFFICE_BACKEND is not set, fall back to SMTP
-    if 'post_office.EmailBackend' in backends['default']:
+    if 'post_office.backends.EmailBackend' in backends['default']:
         backends['default'] = 'django.core.mail.backends.smtp.EmailBackend'
 
     return backends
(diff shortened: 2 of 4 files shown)


selwin:
Let’s use a separate `version.py` file like what we do in RQ so we don’t lose the ability to import from `__init__.py`. This way, we get the best of both worlds :)

selwin:
https://github.com/rq/rq/blob/master/rq/version.py

jrief:
@selwin that unfortunately doesn't work. If `setup.py` for instance calls
```
from post_office.version import VERSION
```
, then `post_office.__init__` is executed anyway, hence
```
from .backends import EmailBackend
```
fails because Django isn't initialized yet.

selwin:
Ah ok, thanks for the explanation.

In that case, I'm leaning towards not merging in this PR because:
1. It's not backwards compatible
2. While it's true that making releases is a bit more tedious for us maintainers, it's not worth sacrificing developer ergonomics for this. `post_office.EmailBackend` is simply better than `post_office.backends.EmailBackend :)

Mogost:
I would like to return to this.
The proposed changes from @jrief  are correct in my opinion.
It should be convenient for everyone, both developers and maintainers.
In my opinion, ideally, we need to add an automatic deployment with CI (travis).
I could work on this, but this requires a fundamental agreement with @jrief and @selwin .

jrief:
@Mogost Thanks for readdressing this. The only solution I see here, which could help us, is to add a file, say `version.txt` and parse it. Adding a file, say `version.py` doesn't function for the reasons explained above.

Mogost:
Ok. What if we leave everything as it is, but use lazy loading.
I am not sure but something like that in __init__.py would work:
```
def EmailBackend:
    from .backends import EmailBackend
    return EmailBackend()
```
@jrief could you check this approach?

jrief:
superseeded by #303

PR #275, opened 2019-09-13
Add attachment inline to Email admin

ilikerobots:
Allow viewing/editing attachments from the Email admin change view


selwin:
Thanks!

PR #277, opened 2019-09-23
Add EMAIL_TEST_MODE settings to allow email redirection in developeme…

Houtmann:
If in the settings.py in django project have 
`EMAIL_TEST_MODE = 'toto@gmail.com'`

All mail are redirect to toto@gmail.com

It is a idea from django-yubin project 
https://django-yubin.readthedocs.io/en/latest/settings.html

### DIFF
--- a/post_office/backends.py
+++ b/post_office/backends.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from email.mime.base import MIMEBase
 from django.core.files.base import ContentFile
 from django.core.mail.backends.base import BaseEmailBackend
+from django.conf import settings
 
 from .settings import get_default_priority
 
@@ -51,6 +52,10 @@ class EmailBackend(BaseEmailBackend):
                 else:
                     attachment_files[attachment[0]] = ContentFile(attachment[1])
 
+            # If EMAIL_TEST_MODE is in settings.py all email are redirect
+            test_email = getattr(settings, 'EMAIL_TEST_MODE', False)
+            if test_email:
+                email_message.to = email
             email = create(sender=from_email,
                            recipients=email_message.to, cc=email_message.cc,
                            bcc=email_message.bcc, subject=subject,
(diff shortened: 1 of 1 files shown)


jrief:
Rather than redirecting an email to another address, I simply would disable the sending queue in testing mode. This can be done by adding some logic to your `settings.py`.
Then you can access the mail-queue through your Django-Admin and check, which mail will be delivered to whom.
From my point of view, such an additional setting may cause more harm than benefit.

Houtmann:
I dont't know how disable sending queue ?
I did this pull request for help developers who want test if smtp work and mail are rendering correctly

jrief:
> I don't know how disable sending queue ?
Either - Don't run it. Emails in django-post_office are send by an ansynchronos daemon.
Or - Configure your SMTP settings pointing to nowhere.

Houtmann:
Yes i'm stupid

jarhainx:
It's a good idea that would be useful to me in development.

When I do developments I like to be able to check the emails and being sure not to use the wrong recipient.

I can help for documentation of this feature if it helps ?

selwin:
I'm not against having a `REDIRECT_ALL_MAIL_TO` option that redirects all emails to the specified address.

However, if you want just want to prevent your app from sending emails out during tests, a better way would be to set `EMAIL_BACKEND` to 'django.core.mail.backends.locmem.EmailBackend'. More info on that [here](https://stackoverflow.com/a/16236241).

Houtmann:
I thought the option was interesting because it protects and allows developers to test the rendering of emails.
But if you went I close this pull request, tell me

selwin:
I think testing the rendering of emails is valid. Let's change the option to `REDIRECT_MAILS_TO`. 

However, I'd like to change the implementation to change the recipient when the email is actually being sent, not when the email is written to the DB.

Tests are also required if we want to merge this feature in :)

selwin:
Or actually, I think `OVERRIDE_RECIPIENTS` would be a better settings name.

Houtmann:
Can i add this logic in the dispatch method in the models ?

Houtmann:
Ok I close this pull request, I will do a new

Issue #171, opened 2016-08-29
retry_failed command needed


subramanyamVemu:
I am from the django-mailer world which has a retry deferred command to add the failed mails in the queue again 

[Source File ](https://github.com/pinax/django-mailer/blob/master/mailer/management/commands/retry_deferred.py)

https://github.com/pinax/django-mailer/blob/master/docs/usage.rst#clear-queue-with-command-extensions

Is an equivalent command applicable to django-post_office

selwin:
We don't have that yet. A PR would be welcome 😃

PR #278, opened 2019-09-25
Add auto retry mail failed mode with requeued mail , and change sheduled time Issue #171

Houtmann:
For answer to issue #171 I created a retry_failed command.

This command replace failed email in queue with a max_retries argument


jrief:
This is a valid use case!

However, instead of creating another task, I would add requeuing [here](https://github.com/ui/django-post_office/blob/5dffcf74226af20096b6ce27b0000e2c2ff02306/post_office/mail.py#L272).
Reason for this change request is, that with your management command, sending an email may happen twice in a very short period, depending on the next call. It also can be disturbing to see status `failed` and a few minutes later `queued`.

Therefore, if sending of an email failed, I would requeue it immediately. Add `STATUS.queued` with a `scheduled_time` in the future. The timedelta for rescheduling and the number of retries should be part of the projects settings, rather than having to specify them on the command line.
This proposed implementation would also hit the database less often.

An additional change would be to introduce another status, say `STATUS.requeued`, but @selwin must decide that.

Houtmann:
Ok I think the job is done

selwin:
I think we should add more granularity as to how emails get retried. For example, I'd want newsletter emails to get retried after failure, but I wouldn't want emails containing one time passwords to get retried 15 minutes later (the password would be worthless by then).

I'd suggest having an interface like:
```python
mail.send(..., max_retries=4, retry_interval=900). # retry_interval could be in seconds or timedeltas
```

Later on, I'd even want to add retrying mechanisms based on failure types, such as only retrying only when encountering network errors and not retrying on email bounces.

jrief:
> but I wouldn't want emails containing one time passwords to get retried 15 minutes later

Good point!
What about implementing the header fields `Expiry-Date` and `Expires` as specified in [RFC4021](https://tools.ietf.org/html/rfc4021#page-30)?
Together with the header field `Priority`, the arguments `max_retries` and `retry_interval` could be computed on the fly by **django-post_office**.
This approach, in my opinion is more standard conform with the RFC.

Anyway, we still would have to change the model and add a `DateTimeField` named `expires_at` to `Email`.  What do you think?

selwin:
>> Anyway, we still would have to change the model and add a DateTimeField named expires_at to Email. What do you think?
I think is a good first step.

The second step would be to add `max_retries` to `Email` model (and perhaps another field to keep track of how many attempts have been made to deliver the email).

jrief:
> The second step would be to add max_retries to Email model 

This could be done through a configuration settings. Then for instance, emails tagged as *high* priority are attempted every 5 minutes, *normal* priority every hour and *low* priority every 12 hours. Together with the expiry timestamp, the number of max attempts is determined organically.

The reason I would do it that way is, that by using the fields `priority` and `expiry`, **django-post_office**'s behaviour then would be standard conformant and similar to user mail clients.
Therefore I would not add `max_retries`, because it tells nothing to the caller of the Post-Office API. The only things relevant to him are the importance of an email and until when it makes sense to be delivered.

> and perhaps another field to keep track of how many attempts have been made to deliver the email

Since the number of delivery attempts can be counted by filtering the log, isn't that redundant?

Houtmann:
Like this ?
`POST_OFFICE = {
    'OVERRIDE_RECIPIENTS': 'hhoutmann@gmail.com',
    'MAIL_PRIORITY_TIME_TO_RETRY': {
        'LOW': datetime.timedelta(hours=4),
        'MEDIUM': datetime.timedelta(hours=1),
        "HIGH": datetime.timedelta(minutes=10),
        "NOW": datetime.timedelta(minutes=1),
    },
    'MAIL_MAX_RETRY': 4,`

jrief:
@Houtmann That would be my preferred solution, but @selwin has to decide.

Maybe, we could use the enums from `post_office.models.PRIORITY`, which however would have to go into a separate file to be importable into `settings.py`.

selwin:
> Since the number of delivery attempts can be counted by filtering the log, isn't that redundant?

Logs may be disabled so we really can't count on logs being present to keep track of retries.

> This could be done through a configuration settings. Then for instance, emails tagged as high priority are attempted every 5 minutes, normal priority every hour and low priority every 12 hours. Together with the expiry timestamp, the number of max attempts is determined organically.

I don't like this approach because I feel like the importance of emails may not directly correspond to whether you want them to be retried. For example, both email types, like those containing one time passwords and suspicious account activities are both high priority, but:
* Emails containing one time passwords may not be retried.
* Emails notifying users of suspicious account activity should absolutely be retried.

So it would be better to be able to configure retries for individual emails.

selwin:
I have created a separate issue for setting email expiration here: https://github.com/ui/django-post_office/issues/281

Email expiration should be handled differently than retries, but they could also be used together.

jrief:
@selwin since this pull request is pending for almost half year, shall I merge it, or do we still wait for improvement?
In my opinion, even if it's not a the best solution, this proposal is a much better what we have now. And such as feature is really useful! Assume configurations where the mail queue is processed every minute, then you really miss the opportunity to resend mail, in case of a short SMTP-server outage.

Houtmann:
I'm looking for time to work on it, at the moment I don't have enough free time.
But I think this evolution is necessary

selwin:
As for the first step, I'm ok with @Houtmann 's original proposal of specifiying global retries:
```python
POST_OFFICE = {
     'MAX_RETRIES': 4
     'RETRY_INTERVAL': datetime.timedelta(minutes=15)
}
```

But having a `email.number_of_retries` field to keep track of how many times an email has been retried is a must have for these reasons:
1. This will cause infinite retries for people who don't have logs enabled
2. The current implementation don't have good performance characteristics, causing O(n) database queries (checking logs) for each email failure. So in the event that your server loses internet connection, it will hammer the DB with repeated requests.

Houtmann:
See #302

PR #280, opened 2019-09-27
Override recipients settings #277

Houtmann:
Related to #277

If in the settings.py in django project have
```python
POST_OFFICE = {
    'OVERRIDE_RECIPIENTS': 'toto@gmail.com'
}```

All mail are redirect to toto@gmail.com
Option is interesting because it protects and allows developers to test the rendering of emails.

It is a idea from django-yubin project
https://django-yubin.readthedocs.io/en/latest/settings.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -512,6 +512,18 @@ setting ``DEFAULT_PRIORITY``. Integration with asynchronous email backends
         'DEFAULT_PRIORITY': 'now'
     }
 
+Override recipient
+----------------
+
+The default is disabled. If you went redirect all sended mails, for testing mail rendering in developpement period
+
+.. code-block:: python
+
+    # Put this in settings.py
+    POST_OFFICE = {
+        'OBERRIDE_RECIPIENTS': 'to@exemple.com'
+    }
+
 Log Level
 ---------
 
--- a/README.rst
+++ b/README.rst
@@ -521,7 +521,7 @@ The default is disabled. If you went redirect all sended mails, for testing mail
 
     # Put this in settings.py
     POST_OFFICE = {
-        'OVERRIDE_RECIPIENTS': 'to@exemple.com'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: post_office/models.py [hunk 1]
```diff
@@ -96,6 +96,10 @@ def prepare_email_message(self):
         Returns a django ``EmailMessage`` or ``EmailMultiAlternatives`` object,
         depending on whether html_message is empty.
         """
+        if get_override_recipients():
+            self.to = [get_override_recipients()]
+            self.save(update_fields=['to'])
```
selwin: This shouldn't be persisted to DB.

README.rst
```diff
@@ -512,6 +512,18 @@ setting ``DEFAULT_PRIORITY``. Integration with asynchronous email backends
...
+Override recipient
+----------------
+
+The default is disabled. If you went redirect all sended mails, for testing mail rendering in developpement period
+
+.. code-block:: python
+
+    # Put this in settings.py
+    POST_OFFICE = {
+        'OVERRIDE_RECIPIENTS': 'to@exemple.com'
```
selwin: In addition to a single email address, I think we should also allow this configuration option to support a list of addresses, such as `['a@example.com', 'b@example.com']`

README.rst
```diff
@@ -512,6 +512,18 @@ setting ``DEFAULT_PRIORITY``. Integration with asynchronous email backends
         'DEFAULT_PRIORITY': 'now'
     }
 
+Override recipient
```
selwin: `Override Recipients`

Houtmann: I don't understand this review

README.rst
Houtmann: Commit #a5dd6d8

post_office/models.py
Houtmann: Commit #a5dd6d8

README.rst
selwin: It's ok, I'll merge this PR in and touch up the docs later.

selwin:
Thanks!

PR #285, opened 2019-12-09
Replace jsonfield with jsonfield2

### LINKED ISSUE #284: External dependency is no longer supported
Incompatible with newer versions of the Django.
Problem with jsonfield
There is some ways to resolve problem.
Separeted from #274 
https://github.com/ui/django-post_office/issues/274#issuecomment-561696438

Mogost:
Fix #284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,15 @@ matrix:
 
     # Python 3.6
     - python: 3.6
-      env: TOXENV=py36-django111,py36-django20,py36-django21,py36-django22
+      env: TOXENV=py36-django111,py36-django20,py36-django21,py36-django22,py36-django30
 
     # Python 3.7
     - python: 3.7
-      env: TOXENV=py37-django20,py37-django21,py37-django22
+      env: TOXENV=py37-django20,py37-django21,py37-django22,py37-django30
 
     # Python 3.8
     - python: 3.8
-      env: TOXENV=py38-django20,py38-django21,py38-django22
+      env: TOXENV=py38-django20,py38-django21,py38-django22,py38-django30
 
     # Django Master
     - python: 3.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Version X.X.X
+-------------
+* Drop support for Django < 1.11 and Python < 3.5.
+* Replace unsupported dependency ``jsonfield`` with supported fork ``jsonfield2``.
+
 Version 3.2.1
 -------------
 * Fix #264: Replace unicode elipsis against 3 dots.
--- a/README.rst
+++ b/README.rst
@@ -19,8 +19,8 @@ Some awesome features are:
 Dependencies
 ============
 
-* `django >= 1.11 <http://djangoproject.com/>`_
-* `django-jsonfield <https://github.com/bradjasper/django-jsonfield>`_
+* `django >= 1.11 <https://djangoproject.com/>`_
+* `jsonfield2 <https://github.com/rpkilby/jsonfield2>`_
 
 
 Installation
--- a/post_office/fields.py
+++ b/post_office/fields.py
@@ -1,3 +1,4 @@
+import django
 from django.db.models import TextField
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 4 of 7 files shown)


Mogost:
@selwin I decided to replace the unsupported library with a fork. This is probably the safest way.

@rpkilby Our build failed only with the latest **jsonfield2** (**3.1.0** version). With **3.0.3** build is successful.
Could you check why the build with the latest version is unsuccessful?
It would also be great to receive from you a confirmation of the intention to continue supporting jsonfield2.
Also, if possible, tell us how safe the transition from **jsonfield** to **jsonfield2** is for us.

Mogost:
I think the build failed due to https://github.com/rpkilby/jsonfield2/issues/20

Mogost:
Oh, sorry. 
Missed in ChangeLog
> Drop Django 2.1 (and below) support
@selwin What do you think about ending support Django 1.11?

Mogost:
Also @jrief what do you think we should do?

rpkilby:
Also, per the 3.0 [release notes](https://docs.djangoproject.com/en/3.0/releases/3.0/#third-party-library-support-for-older-version-of-django), they recommend 3rd party apps to drop support for Django 2.1 and below.

> Also, if possible, tell us how safe the transition from jsonfield to jsonfield2 is for us.

It should be a simple drop-in replacement. Testing locally, I didn't run into any data issues moving from jsonfield to jsonfield2. Make sure that you uninstall jsonfield before installing jsonfield2, since they share the same module paths, and uninstalling jsonfield after the fact will remove modules owned by jsonfield2.

Mogost:
Also, I made a request to transfer access from an unsupported jsonfield to the owner of jsonfield2.
https://github.com/rpkilby/jsonfield2/issues/24
The original owner has replied.
I think we can expect new releases to appear on jsonfield.

selwin:
👍🏻

blorenz:
Here we sit, with no movement on `jsonfield`. Is there an alternative action that can be taken?

selwin:
We can go with `jsonfield2` for the next release. I just want to confirm with @Mogost that the forked library is fully backwards compatible with `jsonfield`. There's no migration step that needs to be done, right?

Mogost:
@selwin No migration step needed.
In the last commit, I froze jsonfield2 on a last compat with Django 1.11 version.
As we see the tests passed.
I think after the official EOL Django==1.11.* we will remove the freeze

selwin:
Ok, let's go with jsonfield2 then. Additionally, I'm ok with supporting django >= 2.0 because keeping compatibility across Django 1.x all the way to 3.x is more trouble than it's worth.

Mogost:
> Ok, let's go with jsonfield2 then. Additionally, I'm ok with supporting django >= 2.0 because keeping compatibility across Django 1.x all the way to 3.x is more trouble than it's worth.

I propose to keep 1.11 until the official EOL which is in April.
By April I will prepare a more meaningful PR which drop support olds django.

selwin:
How about we merge that PR in and release the next version in April? After that gets merged in, we also have to do some code maintenance like removing compatibility codes and decide when to launch it.

I'm actually not too worried about dropping Django 1.11 support because Django themselves recommend it this time around. https://docs.djangoproject.com/en/3.0/releases/3.0/#third-party-library-support-for-older-version-of-django

Mogost:
I would suggest merging this PR now and releasing a new version now too.
All tests that are in the project were successful with all versions of django from 1.11 to django-master.

Mogost:
@selwin Please merge that.

Mogost:
@selwin Thanks! New release is also expected.

rpkilby:
Hi all. I've recently been given ownership of `jsonfield` and have merged `jsonfield2`'s changes back in. `jsonfield` v3.0 is up on PyPI, which is Python 3 and Django 2.2/3.0-compatible.

kakulukia:
I saw the last update and wanted to try it out .. :/ 

File ".../site-packages/post_office/settings.py", line 105, in <module>
    context_field_class = import_string(CONTEXT_FIELD_CLASS)
  File ".../site-packages/django/utils/module_loading.py", line 22, in import_string
    raise ImportError('Module "%s" does not define a "%s" attribute/class' % (
ImportError: Module "jsonfield" does not define a "JSONField" attribute/class

```
django-post-office 3.3.0 
jsonfield2         3.0.3
```

rpkilby:
Hi @kakulukia. Try reinstalling jsonfield2 in your environment. It sounds like you're running into the same issue as https://github.com/dj-stripe/dj-stripe/pull/1023

kakulukia:
Thanks! This worked and should be a general warning for anyone trying to upgrade.

rpkilby:
Fortunately, this should no longer be an issue as projects can now go back to using `jsonfield`.

PR #292, opened 2020-02-12
Add -x, --exit-after-batch arguments to send_queued_mail command

marguslaak:
This feature is useful for email servers that have low throttling values. One can configure crontab with proper time interval and appropriate BATCH_SIZE to make sure that all the emails get sent

### DIFF
--- a/post_office/management/commands/send_queued_mail.py
+++ b/post_office/management/commands/send_queued_mail.py
@@ -34,6 +34,11 @@ class Command(BaseCommand):
             type=int,
             help='"0" to log nothing, "1" to only log errors',
         )
+        parser.add_argument(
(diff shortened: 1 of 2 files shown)


post_office/management/commands/send_queued_mail.py
```diff
@@ -34,6 +34,11 @@ def add_arguments(self, parser):
             type=int,
             help='"0" to log nothing, "1" to only log errors',
         )
+        parser.add_argument(
+            '-x', '--exit-after-batch',
```
selwin: I think this feature makes sense. Can you slightly change this to `--max-batches`? `--max-batches=2` means it will exit after two batches of emails are sent.

PR #296, opened 2020-03-04
Celery integration

jrief:
This pull request adds Celery support without the need to especially configure Post Office for it.
It also adds a signal handler, which can be used by third party apps, whenever an email has been queued.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -477,10 +477,10 @@ something such as:
 	. post_office.tasks.send_queued_mail
 
 
-Emails are now beeing delivered by the Celery worker immediatly after they have been queued. In
-order to make this happen, the project's ``celery.py`` setup shall invoke the autodiscover_tasks_
+Emails will now be delivered by the Celery worker, immediatly after they have been queued. In order
+to make this happen, the project's ``celery.py`` setup shall invoke the autodiscover_tasks_
 function. There is no need to otherwise configure Post Office for integrating with Celery. However,
-in case of a temporary delivery failure, we might want to retry to send those emails by a periodic
+in case of a temporary delivery failure, we might want retrying to send those emails by a periodic
 task. This can be done by a simple `Celery beat configuration`_, for instance through
 
 .. code-block:: python
@@ -505,6 +505,7 @@ month.
 .. _Celery beat configuration: https://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#entries
 .. _Django Celery Beat: https://django-celery-beat.readthedocs.io/en/latest/
 .. _autodiscover_tasks: https://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery.autodiscover_tasks
+
 
 Integration with uWSGI
 ======================
--- a/README.rst
+++ b/README.rst
@@ -478,9 +478,10 @@ something such as:
 
 
 Emails are now beeing delivered by the Celery worker immediatly after they have been queued. In
-order to make this happen, there is no need to otherwise configure Post Office. However, in case
-of a temporary delivery failure, we might want to retry to send those emails by a periodic task.
-This can be done by a simple `Celery beat configuration`_, for instance through
+order to make this happen, the project's ``celery.py`` setup shall invoke the autodiscover_tasks_
+function. There is no need to otherwise configure Post Office for integrating with Celery. However,
+in case of a temporary delivery failure, we might want to retry to send those emails by a periodic
+task. This can be done by a simple `Celery beat configuration`_, for instance through
 
 .. code-block:: python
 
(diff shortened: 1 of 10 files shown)


selwin:
@jrief thanks for the PR. I think this is great, but I'd like to evaluate the additions separately. From eyeballing this PR, there are a few things:

1. The refactoring of `cleanup_expired_mails` into its separate function. This looks good.
2. The addition of Django signals. I think signals should be sent not just when `priority` is `NOW`.
3. Celery integration, this should be moved into its own `contrib.celery` namespace.

jrief:
> 1. The refactoring of cleanup_expired_mails into its separate function. This looks good.

what do you mean with that?

> 2. The addition of Django signals. I think signals should be sent not just when priority is NOW.

Sure, didn't think about that. This is because we want to keep emails with `priority=NOW` inside the request/response cycle. However, what about thirty party apps, listening onto that signal? Wouldn't it be unintuitive, if they wouldn't be informed about `priority=NOW` emails? Other proposal: We pass the priority and some other informations about the emails though the signal handler.

> 3. Celery integration, this should be moved into its own contrib.celery namespace.

The problem here is that the Celery integration for post_office consists of only the file `tasks.py`. This name however is a magic file name for Celery, in that sense that django-celery iterates over all installed apps and looks for a file name `tasks.py`. If we move that file into its own namespace, then we have to add `post_office.contrib.celery` to the project's `INSTALLED_APPS`. I personally would prefer an automatic approach, but I can change that, if you prefer it otherwise.

jrief:
@selwin Since it is considered bad practice to use Django signals for app-internal communication, I replaced that with a direct call. However, I left the signal emitter in case other third party apps might need to be informed about mails in the queue. This implicitly should resolve annotation 2.

selwin:
I'm ok with having `tasks.py` since it doesn't interfere with the core source code. I also don't mind merging in such that `send_queued_mail` and `cleanup_mail` chages that allows them to be used with Celery Beats.

I'm hesitant to merge in changes like [this](https://github.com/ui/django-post_office/pull/296/files#diff-1332baf708f2e43e8811138a41db1644R155) because I find it to be too invasive.

Can we think of a way to create hooks that can be used by third party task queues?

jrief:
> Can we think of a way to create hooks that can be used by third party task queues?

But you agree, that it's OK having a signal triggered, after we put mail into the queue?

> I'm hesitant to merge in changes like this because I find it to be too invasive.

So what else can we use?
* A list of callables in a configuration directive.
* A single callable in a configuration directive.
* A registered signal handler, although it's considered an [anti-pattern](https://lincolnloop.com/blog/django-anti-patterns-signals/). This was my first proposal.

In my opinion, Celery as a task queue is so ubiquitous, that having it as soft dependency does not harm the project.

selwin:
> But you agree, that it's OK having a signal triggered, after we put mail into the queue?

Triggering a signal is still ok for me, since this is the natural use case for signals. I'd perhaps skip sending a signal on `send_many()` though, or change the arguments sent to be a list of emails or email IDs.


> So what else can we use?
> 
> * A list of callables in a configuration directive.
> * A single callable in a configuration directive.
> * A registered signal handler, although it's considered an [anti-pattern](https://lincolnloop.com/blog/django-anti-patterns-signals/). This was my first proposal.
The funny thing is, I think this article actually mentions what we want to accomplish as a good use case for signals.

Django Anti-Patterns: Signals | Lincoln Loop 2020-03-31 21-54-35

I think we should go with your first proposal (i.e signals).

post_office/mail.py
```diff
@@ -163,7 +164,9 @@ def send_many(kwargs_list):
     emails = []
     for kwargs in kwargs_list:
         emails.append(send(commit=False, **kwargs))
-    Email.objects.bulk_create(emails)
+    if len(emails) > 0:
+        Email.objects.bulk_create(emails)
+        email_queued.send(sender=Email, emails=emails)
```
selwin, line 169: Is passing `emails=emails` here ok? The signal definition only mentions `email` as the argument.

README.rst
```diff
@@ -479,8 +446,129 @@ You may want to set these up via cron to run regularly::
...
+.. code-block:: python
+
+	app.conf.beat_schedule = {
+	    'send-queued-mail': {
+	        'task': 'post_office.tasks.send_queued_mail',
+	        'schedule': 600.0,
+	    },
+	}
+
+This will (re-)send queued emails every 10 minutes, which is handy in case the first (and immediate)
```
selwin, line 496: I think we should reword this to `This will send queued emails every 10 minutes.` and not mention resending when delivery has failed.

selwin:
I merged your other PR and that caused conflicts with this one, sorry about that! Could you please fix the conflicts?

Issue #291, opened 2020-02-12
Version not updated in git


Mogost:
https://github.com/ui/django-post_office/blob/6ee240e466762bac0e4713e9042f296dca4798a8/post_office/__init__.py#L1

selwin:
This is fixed.

PR #302, opened 2020-03-30
Add Requeued Mode with sheduled Time

Houtmann:
In response to pull request #278

PR #303, opened 2020-03-31
Prevent the need for updating VERSION on two different places

jrief:
This pull requests superseeds #265 

It shall prevent issues such as #131, #151, #252, #291 which unfortunately occur in regular intervals, due to the fact, that django-post_office keeps the version tuple in two different files.

@Mogost please review, so that we can fix this issue for ever.


post_office/__init__.py
```diff
@@ -1,4 +1,7 @@
-VERSION = (3, 3, 1)
+from os.path import dirname, join
+
+with open(join(dirname(__file__), 'version.txt'), 'r') as fh:
+    VERSION = eval(fh.read())
```
selwin: Just a minor nitpick, can we just read the file and treat it as string and split the version codes by `,`? I try to avoid using eval whenever I can.

jrief: I fully agree that using `eval` is an absolute no-go for untrusted sources. But here we are in full control of the file `version.txt`, so it's a "dangerous"  as putting any code into `__init__.py`, `version.py` or anywhere else.

Parsing the file, means to sort out all kinds of nasty problems, such as interpreting numbers correctly and stripping whitespace. And I'm not sure if that step is optimized properly, so it could be that it will be performed each time one loads any submodule from `post_office`.

Other proposal would be to use [ast.literal_eval](https://docs.python.org/3/library/ast.html#ast.literal_eval):

> Safely evaluate an expression node or a string containing a Python literal or container display. The string or node provided may only consist of the following Python literal structures: strings, bytes, numbers, tuples, lists, dicts, sets, booleans, and None.
This can be used for safely evaluating strings containing Python values from untrusted sources without the need to parse the values oneself. It is not capable of evaluating arbitrarily complex expressions, for example involving operators or indexing.

selwin: Ok, let's use `ast.literal_eval` then.

selwin:
Thanks!

jrief:
@selwin Thanks for merging.
I hope you don't mind, if I'm going to release version 3.4.0? Since celery is a new feature, a minor version jump is appropriate.

selwin:
Ok. Email retries in https://github.com/ui/django-post_office/pull/302 can be released in 3.5.0 then.

PR #308, opened 2020-04-23
Allow tasks.py to be imported if celery is not installed

### LINKED ISSUE #307: Version 3.4.0 broken with Huey
Version 3.4.0 added celery tasks in `post_office.tasks`.

Unfortunately another task queue Huey uses the exact same "find apps' `tasks` submodules" auto-discovery ([docs](https://huey.readthedocs.io/en/latest/contrib.html#running-the-consumer)). This means when using with Huey and auto-discovery on, `post_office` breaks `run_huey` with an `ImportError` for `celery`.

I think other async task runners may have the same issue.

I suggest making the import of celery `try/except` and then only define the tasks if it is imported successfully.

adamchainz:
Fixes #307.

### DIFF
--- a/post_office/apps.py
+++ b/post_office/apps.py
@@ -1,6 +1,5 @@
 from django.apps import AppConfig
 from django.utils.translation import ugettext_lazy as _
-from django.utils.module_loading import import_string
 
 
 class PostOfficeConfig(AppConfig):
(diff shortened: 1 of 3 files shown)


adamchainz:
Your Travis build results aren't coming over to GitHub because GitHub is making them hit the API rate limit: https://jamescooke.info/travis-hitting-githubs-api-limits-for-open-source-projects.html

The ✅ results for this commit 69ca4cb are at https://travis-ci.org/github/ui/django-post_office/builds/678524409

post_office/apps.py
```diff
@@ -1,17 +1,14 @@
...
 class PostOfficeConfig(AppConfig):
     name = 'post_office'
     verbose_name = _("Post Office")
 
     def ready(self):
+        from post_office import tasks
         from post_office.signals import email_queued
 
-        try:
-            queued_mail_handler = import_string('post_office.tasks.queued_mail_handler')
```
jrief, line 14: What about leaving `import_string` as it is? If `celery` can't be imported, then this will fail as well and will be catched by the exception handler below.

This is just a question and not a change request.

jrief:
What about extending `tasks.py` to support Celery and Huey altogether?

PR #314, opened 2020-05-25
Improve admin for email model

jrief:
Still work in progress, but already quite usable.

In the Django Admin for the Email model it
* renders the subject using the provided template, if that exists
* renders HTML and plain text emails, just as they would be delivered by email
* makes many fields read-only, because we shall not modify emails after delivery
* it arranges fields in the list view in a more usable manner


jrief:
Here is a screenshot of the new Django admin for the `Email` model. As one can see, all inlined images are rendered inside the HTML previewer of the Email's detail view.
Screenshot 2020-05-29 at 12 34 56

jrief:
Added unit tests for this feature.
Pull request now is ready for review.

selwin:
Wow, this looks awesome, will look at this PR this weekend :)

jrief:
@selwin Please have a look at this commit: https://github.com/ui/django-post_office/pull/314/commits/f10f3de1e276bac16578590e29063fd7ce0141af

I had to change this part, because while implementing my unit test, I found a bug which presumably never caused any issue. Let me explain:

If you created a `EmailMultiAlternatives` containing just a HTML-only message, then when sending that email, an additional `SafeMIMEText` with content-type `text/plain` was added. Its payload was just the string "body".

I therefore fixed this by using method [walk()](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.walk), which in my opinion is the better way to iterate over multipart alternatives.

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+
+        return fieldsets
+
+    def subject_rendered(self, instance):
+        message = instance.email_message()
+        return format_html('<strong>{}</strong>', message.subject)
+
+    subject_rendered.short_description = _("Subject")
+
+    def render_body_plain(self, instance):
```
selwin, line 172: `render_plaintext_body`

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+        for message in instance.email_message().message().walk():
+            if isinstance(message, SafeMIMEText) and message.get_content_type() == 'text/plain':
+                return format_html('<pre>{}</pre>', message.get_payload())
+
+    render_body_plain.short_description = _("Mail Body")
+
+    def render_body_html(self, instance):
```
selwin, line 179: `render_html_body`

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+        pattern = re.compile('cid:([0-9a-f]{32})')
+        url = reverse('admin:post_office_email_image', kwargs={'pk': instance.id, 'content_id': 32 * '0'})
+        url = url.replace(32 * '0', r'\1')
+        for message in instance.email_message().message().walk():
+            if isinstance(message, SafeMIMEText) and message.get_content_type() == 'text/html':
+                payload = message.get_payload(decode=True).decode('utf-8')
+                body = mark_safe(pattern.sub(url, payload))
```
selwin, line 186: I'm worried that this could cause a stored XSS attack. Does Django have a built in way to mark template strings as safe, except the JS part?

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+                return format_html('<div>{}</div>', body)
+
+    render_body_html.short_description = _("Mail Body")
```
selwin, line 189: `HTML Body`

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+        return format_html('<strong>{}</strong>', message.subject)
+
+    subject_rendered.short_description = _("Subject")
+
+    def render_body_plain(self, instance):
+        for message in instance.email_message().message().walk():
+            if isinstance(message, SafeMIMEText) and message.get_content_type() == 'text/plain':
+                return format_html('<pre>{}</pre>', message.get_payload())
+
+    render_body_plain.short_description = _("Mail Body")
```
selwin, line 177: `Plaintext Body`

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
 
-    to_display.short_description = 'to'
+    to_display.short_description = _("To")
     to_display.admin_order_field = 'to'
 
-
+    def has_add_permission(self, request):
+        return False
+
+    def subject_shortened(self, instance):
```
selwin, line 112: `shortened_subject`

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+
+    def get_fieldsets(self, request, obj=None):
+        fieldsets = [
+            (None, {
+                'fields': ['from_email', 'to', 'cc', 'bcc',
+                           'priority', ('status', 'scheduled_time')],
+            }),
+        ]
+        email_plain, email_html = False, False
+        for message in obj.email_message().message().walk():
```
selwin, line 142: `for part in obj.email_message().message().walk():`

post_office/backends.py
```diff
@@ -28,14 +28,20 @@ def send_messages(self, email_messages):
...
-            message = email_message.body
             headers = email_message.extra_headers
 
-            # Check whether email has 'text/html' alternative
-            alternatives = getattr(email_message, 'alternatives', ())
-            for alternative in alternatives:
-                if alternative[1].startswith('text/html'):
-                    html_message = alternative[0]
+            # Look for first 'text/plain' alternative in email
+            for part in email_message.message().walk():
```
selwin, line 34: Do you mind caching `email_message.message()` to a separate variable so `message()` is not called twice? This method [does quite a few things](https://github.com/django/django/blob/235b68135172c473e022045931719f047cf3dc4a/django/core/mail/message.py#L236) so it'd be more efficient if we cache it.

It would be even better if we can avoid walking over `message()` twice. Something like this perhaps?
```python
message = ''
html_message = ''
for part in email_message.message().walk():
    if part.get_content_type() == 'text/plain':
        message = part.get_payload()
    if part.get_content_type() == 'text/html':
        html_message = part.get_payload()
```

jrief: First, I attempted exactly this approach. The problem is that we then don't know if a message is plaintext, HTML or both. In addition we want to use the first parts, not the last ones. Doing so with only one loop is more complicate.

Moreover, since both loops usually break after one or two iterations, having two loop probably is faster than not breaking and iterating over all attachments.

jrief: fixed

post_office/admin.py
jrief: Good point, this indeed is a possible attack vector. While email clients usually do not execute JavaScript, by adding a `<script>doSomethingEvil();</script>` to the body, it would be executed inside the Django admin.

Django itself does not provide any tools to remove `<script>`-tags. Instead they recommend to use the [bleach](https://pypi.org/project/bleach/) library. I could use this libary, but this would add another third party dependency. So you have to evaluate, if that additional requirement is worthwhile.

I then would also remove the `<style>`-tag. Emails clients ignore it anyway and you can do other nasty stuff with it.

selwin: Well, one thing that's for sure is that we don't want to ship insecure apps.

A good compromise would be to only render the HTML preview if `bleach` is installed. What do you think?

post_office/admin.py
```diff
@@ -64,16 +90,113 @@ class EmailAdmin(admin.ModelAdmin):
...
+        elif has_plaintext_content:
+            fieldsets.append(
+                (_("Text Email"), {'fields': ['render_subject', 'render_plaintext_body']})
+             )
+
+        return fieldsets
+
+    def render_subject(self, instance):
+        message = instance.email_message()
+        return format_html('<strong>{}</strong>', message.subject)
```
selwin, line 168: No need to wrap the subject in `<strong>` tags.

post_office/admin.py
jrief: Instead of bleach, I added the lxml library as an optional dependency. Whenever it is installed, HTML emails are sanitized. Bleach was too strict and required to much special configuration.

README.rst
```diff
@@ -23,6 +23,15 @@ Dependencies
 * `jsonfield <https://github.com/rpkilby/jsonfield>`_
 
 
+Optional Dependency
+-------------------
+
+* `lxml <https://lxml.de/>`_
+
+Add this optional dependency, if you don't trust the HTML beeing rendered inside the Django admin
+backend. This library then removes potentially dangerous tags in order to prevent XSS attacks.
```
selwin, line 32: I think insecure by default is not the right approach here. I'd rather the html body is rendered *only* if lxml is installed. If it isn't, just show a message saying `Install lxml to see HTML preview`.

jrief: After examining which HTML tags, attributes and styles are allowed within emails, I came up with a complete configuration for `bleach`.
Without `bleach`, HTML emails are rendered without tags, using Django's internal `strip_tags`. This setup should be safe regardless of its optional dependencies.

post_office/sanitizer.py
```diff
@@ -0,0 +1,144 @@
...
+            'z-index',
+            'list-style-position',
+            'list-style-tyle',
+        ],
+    ))
+except ImportError:
+    # if bleach is not installed, handle rendered HTML as plain text
+    clean_html = lambda body: mark_safe(format_lazy('<p><em>({heading})</em></p><div>{body}</div>',
+                                                    heading=gettext_lazy("stripping all HTML tags"),
+                                                    body=strip_tags(body)))
```
selwin, line 144: One last comment. If bleach is not installed, we shouldn't preview any HTML email, even after `strip_tags`. We should just make a note that if previes is desired, install bleach. Django's docs [cautions against this](https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.html.strip_tags).

Django Utils | Django documentation | Django 2020-06-16 07-26-46

jrief: Fixed by first stripping tags, then escaping. The other way would create an unreadable mess.

post_office/sanitizer.py
```diff
@@ -0,0 +1,144 @@
...
+            'z-index',
+            'list-style-position',
+            'list-style-tyle',
+        ],
+    ))
+except ImportError:
+    # if bleach is not installed, handle rendered HTML as plain text
+    clean_html = lambda body: mark_safe(format_lazy('<p><em>({heading})</em></p><div>{body}</div>',
+                                                    heading=gettext_lazy("stripping all HTML tags"),
+                                                    body=escape(strip_tags(body))))
```
selwin, line 144: In addition to this, let's display a text informing the user if they want the user to see the nice and properly rendered HTML email, they need to install `bleach`.

If you're putting the body content through `escape`, might as well remove `strip_tags()` so users would see the actualy plain text preview of the HTML email.

jrief: > If you're putting the body content through escape
well, then we don't have to mark the text as safe at all and Django will do it for us.

I tested it with my own generated mails. If you escape everything without stripping HTML tags, you get an unreadable mess of HTML, packed into one line. Therefore stripping first and then escaping at least leaves the text readable, although not very well.

I will a not to install `bleach`.

jrief:
@selwin since all of your change requests have been fixed now, can I merge this?

post_office/sanitizer.py
```diff
@@ -0,0 +1,145 @@
...
+            'list-style-position',
+            'list-style-tyle',
+        ],
+    ))
+except ImportError:
+    # if bleach is not installed, handle rendered HTML as plain text
+    heading = gettext_lazy("Stripping all HTML tags – install 'bleach' to render HTML properly.")
+    clean_html = lambda body: mark_safe(format_lazy('<p><em>({heading})</em></p><div>{body}</div>',
+                                                    heading=heading,
+                                                    body=escape(strip_tags(body))))
```
selwin, line 145: If we're already calling `escape`, HTML tags will be displayed as regular text (e.g `<p>A paragraph</p>`).

Since the purpose of this feature is to give a preview of the rendered content, it'd be better if we lose the `strip_tags` so the HTML source code is properly output.

After `strip_tags` is removed, we can merge this in.

Further improvement to this would be to actually divide the HTML preview into two sections:
1. The actual source code preview
2. The rendered HTML, which only shows if `bleach` is installed.

But I think the improvement can come at a later pull request. This PR has been a long time coming and I also want to see this merged in :).

jrief: Now if `bleach` is installed, emails are rendered as well readable HTML. If bleach is missing, HTML is rendered, but escaped making it hard to read.

> Further improvement to this would be to actually divide the HTML preview into two sections:

I'm not so convinced about this. Users of the Django admin backend do not necessarily have a technical background and "seeing" escaped HTML often does not make sense for them.

selwin:
Thanks for the effort!

PR #315, opened 2020-05-29
Add requeued mode with scheduled- and expired time

### LINKED ISSUE #281: Setting expiration time on emails
You may not want to send certain types of emails after a certain date. Rough ideas on how it should be implemented:
* Add `expiration_time` field on `Email`
* When email is picked up for processing in [send_queued()](https://github.com/ui/django-post_office/blob/master/post_office/mail.py#L181), if `expiration_time` is in the past, change the status to `expired`

jrief:
This pull request superseeds #302 .
It fixes a few issues I found there and it adds an optional `expired_at` timestamp, so that
emails will never be delivered later than that. This has been discussed in #281 and makes a lot of sense together with requeuing. 

Fixes #281
Closes #302


post_office/mail.py
```diff
@@ -171,22 +173,28 @@ def send_many(kwargs_list):
...
+     - Has expires_at after the current time or is None
     """
-    return Email.objects.filter(status=STATUS.queued) \
-        .select_related('template') \
-        .filter(Q(scheduled_time__lte=now()) | Q(scheduled_time=None)) \
-        .order_by(*get_sending_order()).prefetch_related('attachments')[:get_batch_size()]
+    now = timezone.now()
+    query = (
+        (Q(status=STATUS.queued) | Q(status=STATUS.requeued)) &
+        (Q(scheduled_time__lte=now) | Q(scheduled_time__isnull=True)) &
```
selwin, line 184: Nitpick: `Q(scheduled_time=None)`

post_office/mail.py
```diff
@@ -171,22 +173,28 @@ def send_many(kwargs_list):
...
+        (Q(expires_at__gt=now) | Q(expires_at__isnull=True))
```
selwin, line 185: Nitpick: `Q(expires_at=None)`

post_office/mail.py
```diff
@@ -268,12 +278,27 @@ def send(email):
...
+    )
+    query = Q(retry_counter__lte=get_max_retries())
+    num_failed = emails_failed.exclude(query).update(
+        status=STATUS.failed
+    )
+    num_requeued = emails_failed.filter(query).update(
+        number_of_retries=F('retry_counter'),
+        status=STATUS.requeued,
+        scheduled_time=timezone.now() + get_time_delta_to_retry(),
+    )
```
selwin, line 301: I'm not a big fan of this code block because it makes two separate DB queries. Since Django 1.8 is EOL'ed, I think we should use `bulk_update()` (available in Django >= 2.2), the resulting code should also be more concise. Something like this:

```python
max_retries = get_max_retries()
for email in failed_emails:
    if email.number_of_retries < max_retries:
        email.status = STATUS.requeued
    # ...

Email.objects.bulk_update(failed_emails, ['status', 'number_of_retries']
```

post_office/models.py
```diff
@@ -57,6 +58,9 @@ class Email(models.Model):
     last_updated = models.DateTimeField(db_index=True, auto_now=True)
     scheduled_time = models.DateTimeField(_('The scheduled sending time'),
                                           blank=True, null=True, db_index=True)
+    expires_at = models.DateTimeField(_('Expires timestamp for email'),
```
selwin, line 61: `Email won't be sent after this timestamp`

post_office/mail.py
jrief: I believe that the proper way doing it in Django is using [...__isnull=True](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#isnull) otherwise their documentation would state that. The only mentioned alternative is [...__exact=None](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#exact), but then I really prefer `...__isnull=True`.

post_office/mail.py
jrief: detto

post_office/mail.py
jrief: I changed that to a `bulk_update` logic.

post_office/models.py
jrief: done

post_office/mail.py
Mogost: I agree with the nitpick from @selwin.  I always use this approach and it is actually pretty readable.

jrief: @Mogost , this might be true, but then I would like to know why it isn't mentioned anywhere in the documentation. From reading the docs, the `...__isnull=True`, seems to be the right approach.
Or I didn't read the docs properly. Do you have other sources?

Mogost: I only have django documentation. But I will try to convince you that the offer is good.
First of: 
> As a convenience when no lookup type is provided (like in Entry.objects.get(id=14)) the lookup type is assumed to be **exact**.

Second of:
> Exact match. If the value provided for comparison is None, it will be interpreted as an SQL NULL (see isnull for more details).

Therefore, what you mention as an alternative option can be finalized by taking into account the above.

Ultimately, this is just nitpicking.

jrief: Yes, for datetime fields doing the query with `Q(scheduled_time=None)` gives exactly the same compiled SQL as `Q(scheduled_time__isnull=True)` does. But this is not true for all field types, for instance `JSONField`, where `fieldname=None` translates to Javascript Object Notation's `null`, where as `fieldname__isnull=True` translates to SQL's `IS NULL`.

Moreover, while `Q(scheduled_time=None)` may give the desired result, if one has to reverse the logic, `Q(scheduled_time!=None)` doesn't work for obvious reasons. So we have two way of spellings, `Q(scheduled_time=None)` to compile into `IS NULL` and `Q(scheduled_time__isnull=False)`  to compile into `IS NOT NULL`.

Don't you think that this makes things less explicit? Presumably that's the reason, why the official Django docs doesn't encourage its usage.

jrief:
@selwin Can I merge this now? I think that all requested changes have been solved.

selwin:
Please give me a few days to review this again. Sorry for the wait!

post_office/mail.py
```diff
@@ -268,12 +278,28 @@ def send(email):
...
     Email.objects.filter(id__in=email_ids).update(status=STATUS.sent)
 
-    email_ids = [email.id for (email, e) in failed_emails]
-    Email.objects.filter(id__in=email_ids).update(status=STATUS.failed)
+    # Update statuses and conditionally requeue failed emails
+    num_failed, num_requeued = 0, 0
+    max_retries = get_max_retries()
+    scheduled_time = timezone.now() + get_time_delta_to_retry()
+    email_ids = [email.id for email, _ in failed_emails]
+    emails_failed = Email.objects.filter(id__in=email_ids)
```
selwin, line 290: Do we need to re-fetch these emails from the DB? Can't we just use `failed_emails`?

post_office/models.py
```diff
@@ -108,19 +112,25 @@ def prepare_email_message(self):
             html_message = self.html_message
 
         connection = connections[self.backend_alias or 'default']
+        if isinstance(self.headers, dict) or self.expires_at:
+            headers = dict(self.headers or {})
+            if self.expires_at:
+                headers.update({'Expires': self.expires_at.strftime("%a, %-d %b %H:%M:%S %z")})
+        else:
+            headers = None
```
selwin, line 120: Can we simply this to:
```python
if self.expires_at:
    headers = dict(self.headers or {})
    headers.update({'Expires': self.expires_at.strftime("%a, %-d %b %H:%M:%S %z")})
else:
    headers = self.headers
```

Do you think it's better if we add this header during delivery, or whether we should add `Expires` to `self.headers` when email is first created? I have no strong opinion on this, but perhaps adding the header to be stored in `headers` is a bit more explicit.

post_office/settings.py
```diff
@@ -100,6 +104,14 @@ def get_override_recipients():
     return get_config().get('OVERRIDE_RECIPIENTS', None)
 
 
+def get_max_retries():
+    return get_config().get('MAX_RETRIES', 0)
+
+
+def get_time_delta_to_retry():
```
selwin, line 111: `get_retry_timedelta()`

post_office/mail.py
jrief: Yes indeed, `emails_failed` does not have to be a queryset, it can also be a list of objects.

post_office/models.py
jrief: I would add the `Expires` header just before being sent. Reason is, that we use that field to filter our queryset. Remember, emails never are sent after they expired. 
If we put the "expired at" timestamp into the header, we must eventually update it on two different places. If we just use the "Expired" entry in the header, we must filter against a JSON key, which is not very well supported.

selwin:
Thanks!

oesah:
When can we expect a new release on PyPi? I think a lot of people are having issues with `SMTPServerDisconnected` and `please run connect() first` errors. Hopefully, this fixes it.

PR #318, opened 2020-07-01
Keep track of Message-ID

jrief:


jrief:
According to [RFC2822 each email shall contain a Message-ID](https://tools.ietf.org/html/rfc2822#section-3.6.4).
Therefore, Django adds such a Message-ID to each email, if it has not been added to the email headers yet. This ID is very useful when tracking emails. In Google Mail for instance, one can [search for an email](https://support.google.com/mail/answer/7190?hl=en), by entering
```
rfc822msgid:9759.329231145.62086@example.com
```
into the search field.

Up to now, Django generated this Message-ID on the fly, but that information then was lost.
In the next version, **django-post_office** generates this Message-ID itself and stores it as separate
field for each email. This then should help to follow up emails.

This is my last pull request for version 3.5. If accepted, I will build and release django-post_office-3.5.0 as soon as possible.

README.rst
```diff
@@ -630,6 +630,28 @@ Defaults to ``None``. This option is useful if you want to redirect all emails t
...
+the Message-ID consists of two parts separated by the ``@`` symbol: The left part is a generated pseudo random number.
+The right part is a constant string, typically denoting the domain name of the sending entity.
+
+It shall be configured using
+
+.. code-block:: python
+
+    # Put this in settings.py
+    POST_OFFICE = {
+        'MESSAGE_ID_RIGHT': 'example.com'
```
selwin, line 648: I think we can auto generate message ID without having this setting. We can look at the domain part of the `from` address and just use that. One less settings to configure and get wrong :)

jrief: If unset, Django uses the local host name.

The standard states:

> Message-ID consists of a long string of characters (unique identifier of this message on the mail server) ending with the FQDN of the mail server (e.g. 25eb7e48-2252-48ea-80ce-cf2f6119a8e3@ENV95-E2013-1.domain95.lab). Message-ID of an email can be found in the message header. When using CodeTwo software, you can also find this identifier in the software's log files.

Therefore, I wouldn't use the domain part of the `From:` address, because it's not the FQDN of the mail server.
If unconfigured, post_office falls back to the behaviour as Django itself.

selwin: Ok. If that's the case, let's rename this to `MESSAGE_ID_FQDN` which is much more descriptive. Also, should message ID be generated for all emails, or just those with `MESSAGE_ID_FQDN` set? I propose the latter so only those who need it can turn it on.

For my production use case which has millions of emails in the DB, this will use a fair bit of space, so I'll take all the savings that I can get.

jrief: The generation of a Message-ID now is optional. It can be activated using a configuration directive.

post_office/migrations/0010_message_id.py
```diff
@@ -0,0 +1,38 @@
...
+    if get_message_id_on_delivery():
+        return
+    msg_id_fqdn = get_message_id_fqdn()
+    Email = apps.get_model('post_office', 'Email')
+    for email in Email.objects.using(schema_editor.connection.alias).filter(message_id__isnull=True):
+        if email.status in [STATUS.queued, STATUS.requeued]:
+            # create a unique Message-ID for all emails which have not been send yet
+            randint1, randint2 = random.getrandbits(64), random.getrandbits(16)
+            email.message_id = '<{}.{}.{}@{}>'.format(email.id, randint1, randint2, msg_id_fqdn)
+            email.save()
```
selwin, line 17: I'm not sure we should do a forwards migration in this case. This will cause huge spikes in DB writes when migrating DBs with a large number of emails. I think we should just drop this and leave the migration part optional for people who need it.

post_office/test_settings.py
```diff
@@ -37,6 +37,8 @@
         'connection_tester': 'post_office.tests.test_mail.ConnectionTestingBackend',
     },
     'MAX_RETRIES': 2,
+    'MESSAGE_ID_ON_DELIVERY': False,
```
selwin, line 40: No need to have this option, this just complicates things. If FQDN is specified, just generate this when email is initially created.

post_office/test_settings.py
```diff
@@ -37,6 +37,8 @@
...
+    'MESSAGE_ID_FQDN': 'post_office.org',
```
selwin, line 41: We should rename this to `example.org` in case people copy this by mistake.

post_office/migrations/0010_message_id.py
jrief: well, it is optional. If you don't set the configuration directive `MESSAGE_ID_ON_DELIVERY` to `False`, then nothing happens.

post_office/test_settings.py
jrief: I don't share this opinion. A variable shall be responsible for one type of information and only that one.
If we use `MESSAGE_ID_FQDN` then this variable turns on/off Message-ID generation *and* defines the right part of the Message-ID. This is kind of magic, tri-state behavior and should be avoided.

In addition to this, what if a user want to mimic the behavior of Django and use the server's DNS name?
What happens, if Django introduces its own overwritable "right" part of the Message-ID?

Therefore, if you really don't want two variables, let's drop `MESSAGE_ID_FQDN` instead of `MESSAGE_ID_ON_DELIVERY`.

post_office/test_settings.py
jrief: OK

post_office/test_settings.py
selwin: No, what I'm saying is that whether or not the ID is generated on delivery or when the email is created does not matter, since it's just a random uuid.

post_office/mail.py
```diff
@@ -40,6 +42,10 @@ def create(sender, recipients=None, cc=None, bcc=None, subject='', message='',
         bcc = []
     if context is None:
         context = ''
+    if get_message_id_on_delivery():
+        message_id = None
+    else:
+        message_id = make_msgid(domain=get_message_id_fqdn())
```
selwin, line 48: We can store the message_id without the domain part, so this can save a bit of space. Since it's a UUID, we can already uniquely identify an email without the domain part.

The FQDN part can be added when the mail is being delivered.

jrief:
> We can store the message_id without the domain part, so this can save a bit of space.

I wouldn't do that. The Message-ID is a unique identifier as a whole. If ever someone changes this in its settings, then this change should be reflected in the field as well. It would confuse users of that library.

If space is a real concern, then how about making the configuration directive 
`MESSAGE_ID_FQDN` a callable? It then could be called twice, the first time when the Email is put in the queue and the second time, when delivered. We can pass a Boolean to that callable indicating that. This would also remove the need for the configuration directive`MESSAGE_ID_ON_DELIVERY`. What do you think?

jrief:
I hope this now is a feasible compromise:

The Message-ID is generated by default during Email creation and stored in the database field `message_id`. This probably is most common use-case for >90% of all Post-Office users.

The domain part of the Message-ID can optionally be overridden using the settings directive `POST_OFFICE['MESSAGE_ID_FQDN']`.

This feature can explicitly be disabled by setting `POST_OFFICE['MESSAGE_ID_FQDN'] = None`

post_office/test_settings.py
selwin, line 40: This can be deleted since `MESSAGE_ID_ON_DELIVERY` is no longer used.

selwin:
I don't think Message-ID should be generated by default. This is something that users typically don't have to worry about on the application level and is taken care of by the ESP. For example, Mailgun automatically generates a Message-ID for you if you don't provide it https://help.mailgun.com/hc/en-us/articles/360012289953-How-to-Find-a-Message-ID .

I'd require that `Message-ID` generation is off by default. How do we turn it on? I see two possibilities of doing this:
1. If `MESSAGE_ID_FQDN` is set, `Message-ID` generation will be turned on
2. Have an extra boolean `MESSAGE_ID_ENABLED` parameter that governs this behavior.

I'm ok with both options 1 and 2.

jrief:
@selwin I went with option 2:

> Have an extra boolean MESSAGE_ID_ENABLED parameter that governs this behavior.

jrief:
@selwin is my latest change ready to be merged now?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 3.5.0
+-------------
+* Add ``Message-ID`` to the Email models and use it when sending emails. It replaces the
```
selwin, line 6: Do you mind removing the part where this removed ID column in Django's admin?

post_office/admin.py
```diff
@@ -83,7 +83,8 @@ def requeue(modeladmin, request, queryset):
 
 @admin.register(Email)
 class EmailAdmin(admin.ModelAdmin):
-    list_display = ['id', 'to_display', 'shortened_subject', 'status', 'last_updated', 'scheduled_time', 'use_template']
+    list_display = ['truncated_message_id', 'to_display', 'shortened_subject', 'status', 'last_updated', 'scheduled_time', 'use_template']
+    readonly_fields = ['message_id']
```
selwin, line 87: There are two `readonly_fields` definition, here and in line 88. They should be combined.

post_office/migrations/0010_message_id.py
```diff
@@ -0,0 +1,39 @@
...
+    if not get_message_id_enabled():
+        return
+    msg_id_fqdn = get_message_id_fqdn()
+    Email = apps.get_model('post_office', 'Email')
+    for email in Email.objects.using(schema_editor.connection.alias).filter(message_id__isnull=True):
+        if email.status in [STATUS.queued, STATUS.requeued]:
+            # create a unique Message-ID for all emails which have not been send yet
+            randint1, randint2 = random.getrandbits(64), random.getrandbits(16)
+            email.message_id = '<{}.{}.{}@{}>'.format(email.id, randint1, randint2, msg_id_fqdn)
+            email.save()
```
selwin, line 18: I don't think we should need to provide a forward migration to backfill old emails.

CHANGELOG.md
jrief: Sorry, I don't understand what you mean by that.

post_office/admin.py
jrief: Ups, missed that.
Anyway, the Message-ID now is only rendered if available.

post_office/migrations/0010_message_id.py
jrief: This migration only applies if the `MESSAGE_ID_ENABLED` settings is `True` **and only** for unsent emails. I believe, most users would expect this behaviour.

CHANGELOG.md
jrief: Ok. Changed that.

selwin:
Thanks!

jrief:
Thanks for reviewing all this.

I then will release a new version. There are other documented features, which are not part of the latest release.

selwin:
Thanks for your efforts!

Win~

PR #320, opened 2020-07-03
added EmailTemplate.label attribute

madEng84:
added EmailTemplate.label attribute to display an optional human readable name of an EmailTemplate

Enabling this feature, end users will more easily understand which EmailTemplate to use


jrief:
why aren't the `name` and the `description` fields not enough in your use-case?

madEng84:
The 'name' is machine-readable and it should be not editable after creation to avoid bugs in your code. I'm afraid that end user wants to change it to set it more human readable without a "don't touch me" help_text
The 'description' is not concise, it explains with more words the boundary conditions, it explains why but expecially when your code uses it.
An alternative to add a 'label' could be this one: set a forms.select as the widget of the name attribute showing the end user only the human readable names.
We can think to set a setting EMAIL_TEMPLATE_NAME_CHOICES,  documenting it properly, to avoid end users to change it and dev users to misspell it.

What do you think about?

jrief:
I don't see any benefit from this, except adding another moving part to the project and thus another potential source of error.
Apart from that, if we add a `label`-field, then that field shall be editable on the translated templates as well – don't you think so?

Maybe I'm a bit naive here. Could you please give me a concrete example to which values you would set `name`, `label` and `description`.

jrief:
OK, I see your point. So, you need a kind of identifier, so that you can find and access an `EmailTemplate` from your Python code. This would be a valid use-case.

How about leaving the `name`-field as a human readable label and instead adding an optional `identifier` which is language independent?

jrief:
@madEng84 please have a look at pull request #324. There you can use a language independent `identifier` field to access the template. There the `name` field remains human readable.

PR #322, opened 2020-07-09
Fix #321: Render Email Templates in proper locale settings

### LINKED ISSUE #321: Emails are rendered in wrong locale
How to reproduce:

- In Django's Admin backend create an Email Template object.
- To that template add another template in a different language.
- Inside that template add `{{ date }}`.
- Send an email using a `context` containing say `{'date': timezone.datetime(2020, 5, 18)}` using
  `send(recipients=['me@example.com'], sender='from@example.com', template=template, context=context)`
  with the translated template, we have created in step 2.

We expect that `date` is rendered in the locale of the translated template. However it is rendered in the current locale or in the default locale.

jrief:
This fixes #321  (details are explained there)

### DIFF
--- a/post_office/mail.py
+++ b/post_office/mail.py
@@ -7,6 +7,7 @@ from django.db import connection as db_connection
 from django.db.models import Q
 from django.template import Context, Template
 from django.utils import timezone
+from django.utils.translation import override as translation_override
 
 from .connections import connections
 from .models import Email, EmailTemplate, Log, PRIORITY, STATUS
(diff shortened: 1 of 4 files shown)


post_office/mail.py
```diff
@@ -58,14 +59,17 @@ def create(sender, recipients=None, cc=None, bcc=None, subject='', message='',
...
-        _context = Context(context or {})
-        subject = Template(subject).render(_context)
-        message = Template(message).render(_context)
-        html_message = Template(html_message).render(_context)
+        else:
+            language = ''
+        with translation_override(language):
+            _context = Context(context or {})
+            subject = Template(subject).render(_context)
+            message = Template(message).render(_context)
```
selwin, line 71: `post-office` shouldn't implicitly translate the template itself. The translation should be explicitly defined by the user during the creation of the email template.

Many languages that are based on english alphabets (e.g Indonesian) often use mixed english/Indonesian content on purpose for marketing reasons.

PR #333, opened 2020-08-31
Fix #326: get_queued() shall use select_for_update(skip_locked=True)

### LINKED ISSUE #326: Celery Task triggered by signal is not thread safe: Multiple E-Mails sent!

Now the signal is triggered:

```

```

It calls this task here:

```
        """
        """
        send_queued_mail.delay()
```

fadedDexofan:


jrief:
I fully agree that action is necessary for refactoring parts of the code. However, you never shall mix such code style fixes with a functional change. Adding `select_for_update()` is a huge functional change. Please create a pull request on its own, for functional and style changes. This helps reviewers to separate their concerns. Thanks.

fadedDexofan:
@jrief https://github.com/ui/django-post_office/pull/338

PR #334, opened 2020-09-14
Add setting for celery.

elineda:
As celery doesn't allow to not ignore a task, Put a config for disabled celery even if celery is used.

[+] Add a config for get celery enabled or not (default True)
[+] If celery is not enabled in postoffice settings don't import tasks for postoffice

### DIFF
--- a/post_office/settings.py
+++ b/post_office/settings.py
@@ -80,7 +80,7 @@ def get_batch_size():
 
 
 def get_celery_enabled():
-    return get_config().get('CELERY', True)
+    return get_config().get('CELERY_TASKS_ENABLED', True)
 
 
 def get_threads_per_process():
--- a/post_office/settings.py
+++ b/post_office/settings.py
@@ -79,6 +79,10 @@ def get_batch_size():
     return get_config().get('BATCH_SIZE', 100)
 
 
+def get_celery_enabled():
+    return get_config().get('CELERY', True)
+
+
 def get_threads_per_process():
     return get_config().get('THREADS_PER_PROCESS', 5)
 
--- a/post_office/settings.py
+++ b/post_office/settings.py
@@ -80,7 +80,7 @@ def get_batch_size():
 
 
 def get_celery_enabled():
-    return get_config().get('CELERY_TASKS_ENABLED', True)
+    return get_config().get('CELERY_ENABLED', True)
 
 
 def get_threads_per_process():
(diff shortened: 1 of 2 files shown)


post_office/settings.py
```diff
@@ -79,6 +79,10 @@ def get_batch_size():
     return get_config().get('BATCH_SIZE', 100)
 
 
+def get_celery_enabled():
+    return get_config().get('CELERY', True)
```
jrief, line 83: naming this configuration just `CELERY` is certainly a bad idea, since you just want to enable/disable it for `django-post_office`. If someone doesn't like Celery at all, those tasks are never triggered.

post_office/tasks.py
```diff
@@ -5,28 +5,32 @@
...
-    @shared_task(ignore_result=True)
-    def send_queued_mail(*args, **kwargs):
-        send_queued()
 
-    def queued_mail_handler(sender, **kwargs):
-        """
-        To be called by post_office.signals.email_queued.send()
-        """
-        send_queued_mail.delay()
+if get_celery_enabled():
```
jrief, line 15: This can be refactored into the `try`-block, without needing another level of indentation.

jrief (changes requested):
Please note that the final decision about this feature will be made by @selwin

post_office/tasks.py
elineda: Where do you want tu pot it ? We can't put an elif after an except. We will have a indentation level even if I put it into the else

My logic is if celery is not enabled  we do nothings even the import statement.

post_office/settings.py
elineda: image

As the configuration is into the postoffice item, it wil not interfer with celery global configuration.
But I can rename it at "CELERY_TASKS_ENABLED"

post_office/tasks.py
jrief: this would keep the current indentation level:

```python
try:
    if get_celery_enabled():
    else:
        raise NotImplementedError()
except (ImportError, NotImplementedError):
    pass
else:
    ...
```

post_office/settings.py
jrief: Yes, `CELERY_TASKS_ENABLED ` is more meaningful.

jrief (approved):

post_office/settings.py
```diff
@@ -79,6 +79,10 @@ def get_batch_size():
     return get_config().get('BATCH_SIZE', 100)
 
 
+def get_celery_enabled():
+    return get_config().get('CELERY_TASKS_ENABLED', True)
```
selwin, line 83: Can you rename this to `CELERY_ENABLED`?

elineda: Done

PR #337, opened 2020-09-28
Minor fixes to prevent code breaking

patroqueeet:
why that: because otherwise pdf attachments will not be sent properly. they require mimetype and attachment.name given.


jrief (approved):

selwin:
Thanks!

Issue #330, opened 2020-08-17
Upgrading post office to 3.4.1 causes multiple emails to send


himynameistimli:
Hi,

I'm having an issue that appeared after we upgraded to post office 3.4.1.

In our setup, we're using Django 3 and celery for sending emails after a certain period of time (we wait for additional information before deciding to send or cancel the emails). After a period of time, a custom task is called, which calls the `send_queued_mail` function.

With the upgrade to 3.4.1 I noticed that send_queued_mail was being called early (by itself) and multiple times, so my logs showed that emails were sent multiple times instantaneously. This occurred even when I didn't call my own task to call `send_queued_mail`.

It appears to be indeterministic, meaning it happened only happens occasionally.

Downgrading to 3.3.1 solved my problem, and the duplicated emails went away.

fadedDexofan:
https://github.com/ui/django-post_office/pull/333

jrief:
Did you limit the number of simultaneous connections to your SMTP server to 1?

selwin:
Duplicate of https://github.com/ui/django-post_office/issues/326 . See my comment [here](https://github.com/ui/django-post_office/pull/338#issuecomment-719905372). The TLDR is that the celery tasks should acquire a lock before sending emails.

PR #338, opened 2020-10-05
Fix #326: get_queued() shall use select_for_update(skip_locked=True)

### LINKED ISSUE #326: Celery Task triggered by signal is not thread safe: Multiple E-Mails sent!

fadedDexofan:


jrief (approved):

jrief:
@fadedDexofan did this fix your problem of sending the same email multiple times?

fadedDexofan:
@jrief Yep. Use my fork with fix on work for about month. Works fine. Each `mail.send` invoke dispatches task. More mails -> more tasks. First tasks selects with lock all queued mails, rest of all tasks get 0 emails from queue

selwin:
I know this may not be easy, but is there a way to write an automated test for this? This is not an ORM behavior I'm familiar with.

I did a little bit of Googling around, from what I read is that `transaction.atomic` [does not automatically lock selected rows](https://forum.djangoproject.com/t/do-transaction-atomic-locks-db/2515/2).

fadedDexofan:
@selwin `select_for_update` inside transaction block is used to lock rows

jrief:
@selwin shouldn't we merge this even without a test? Reading the issue tracker, this for some users is a huge embarrassment. I then would release the next major version of Post-Office.

selwin:
@jrief if you're referring to issues https://github.com/ui/django-post_office/issues/326 and https://github.com/ui/django-post_office/issues/330 , this will not fix that.

The celery task would need to acquire locks before dispatching the emails, similar to the `send_queued_mail` management command.

post office allows you to specify how much server resource you want to dedicate to deliver emails through `--processes` and `THREADS_PER_PROCESS` option. Not acquiring a lock would break this.

Consider this scenario; you have 10 million emails scheduled for delivery tomorrow. Assuming that you can send 50k emails/minute (most ESPs impose a rate limit), you'll need 200 minutes to finish sending all your emails. If celery task is called once per minute, even with `select_for_update`, by the tenth minute you'll have 10 different processes sending the emails, using 10 times the resources you provisioned for email delivery. This could be bad news for high traffic systems.

This fix is only useful for people who want to deliver emails manually through scripts.

jrief:
OK.
One of the problems with Celery is, that it doesn't prevent reentrancy. In one of my other projects using Celery with long running tasks, I stopped using Celery-Beat. Instead I added something like this:

```python
app = Celery('my_app', broker='redis://localhost')
app.config_from_object(settings, namespace='CELERY')
app.autodiscover_tasks()


@app.task
def do_something_periodically(**kwargs):
    try:
        elapsed_time = -time.monotonic()
        # do something which takes a lot of time
        elapsed_time += time.monotonic()
        logger.info("Elapsed time: {:.2f} seconds".format(elapsed_time))
    finally:  # reschedule task in 60 seconds
        do_something_periodically.apply_async(countdown=60)

do_something_periodically.apply_async(countdown=60)  # trigger for the first time
```

By doing so, my task `do_something_periodically` never runs twice at the same time.

Shall we use this approach in Post-Office as well? It could be optimized by rescheduling immediately if more mail is in the queue, otherwise reschedule in `x` seconds.

selwin:
Our management command already handles this [using lockfile](https://github.com/ui/django-post_office/blob/master/post_office/management/commands/send_queued_mail.py#L42). The celery task should try to acquire the same lock before running.

That way, whether the task is run via cron (every minute) or celery tasks, only one instance of `send_queued()` will be running at any given time.

After each run, the command also checks whether there are more queued emails and if it does, it [executes again](https://github.com/ui/django-post_office/blob/master/post_office/management/commands/send_queued_mail.py#L56) without waiting for the next minute. So you can use a smaller batch size to save some memory, yet it won't stop sending if you have a large amount of email that's queued.

If you want, you can refactor out lines 42-60 from the management command into a separate function that can be reused inside your celery task.

rob101:
@selwin, is the correct approach not to apply this fix but instead to refactor the [send_queued_mail ](https://github.com/ui/django-post_office/blob/c37ff1155eeebd4796fcb56c772ae6ad101bba8a/post_office/tasks.py#L24)task?

```
try:
except ImportError:
    pass
else:
    @shared_task(ignore_result=True)
    def send_queued_mail(*args, **kwargs):
        try:
            with FileLock(default_lockfile):

                while 1:
                    try:
                        send_queued()
                    except Exception as e:
                        logger.error(e, exc_info=sys.exc_info(),
                                     extra={'status_code': 500})
                        raise

                    # Close DB connection to avoid multiprocessing errors
                    connection.close()

                    if not Email.objects.filter(status=STATUS.queued) \
                            .filter(Q(scheduled_time__lte=now()) | Q(
                        scheduled_time=None)).exists():
                        break
        except FileLocked:
            logger.info('Failed to acquire lock, terminating now.')
```

selwin:
@rob101 correct.

rob101:
@selwin Could this also be solved by simply limiting celery concurrency to 1 for the relevant worker? E.g. `celery -A settings worker -l info --concurrency 1`

selwin:
It won’t be a clean fix from post office’s perspective, because you’re relying on the user to configure things to make it work properly, while it should work out of the box.

If a PR to fix this is not merged within a few weeks, I think I’d make a release to disable this feature temporarily as this bug could lead to monetary (and reputation) damages of sending multiple copies of the same email.

patroqueeet:
> @selwin, is the correct approach not to apply this fix but instead to refactor the [send_queued_mail ](https://github.com/ui/django-post_office/blob/c37ff1155eeebd4796fcb56c772ae6ad101bba8a/post_office/tasks.py#L24)task?
> 
> ```
> try:
>     from celery import shared_task
> except ImportError:
>     pass
> else:
>     @shared_task(ignore_result=True)
>     def send_queued_mail(*args, **kwargs):
>         try:
>             with FileLock(default_lockfile):
> 
>                 while 1:
>                     try:
>                         send_queued()
>                     except Exception as e:
>                         logger.error(e, exc_info=sys.exc_info(),
>                                      extra={'status_code': 500})
>                         raise
> 
>                     # Close DB connection to avoid multiprocessing errors
>                     connection.close()
> 
>                     if not Email.objects.filter(status=STATUS.queued) \
>                             .filter(Q(scheduled_time__lte=now()) | Q(
>                         scheduled_time=None)).exists():
>                         break
>         except FileLocked:
>             logger.info('Failed to acquire lock, terminating now.')
> ```

when I use this logic inside my app and remove the celery beat entry for the native task, what about the signal handler? this still calls the wrong task... can we make this either properly released or at least configurable?

patroqueeet:
one more: sry for bugging: the `connection.close()` kills the test suite because tests run with `CELERY_ALWAYS_EAGER`(there is no celery in CI) and then in this process db connection for all tests is closed. added a check for this locally:

```
 87                 # Close DB connection to avoid multiprocessing errors
 88                 if not settings.CELERY_TASK_ALWAYS_EAGER:
 89                     connection.close()
```

selwin:
I think you should always close the DB connection regardless of Celery settings. Whenever you make a query using Django's ORM, if the connection is closed, it will always open a new one.

patroqueeet:
> I think you should always close the DB connection regardless of Celery settings. Whenever you make a query using Django's ORM, if the connection is closed, it will always open a new one.

@selwin I understand what you're saying. 

Symptoms were: connection close added to code, push to CI resulted in hundreds of tests failing with `connection already closed`. After adding the conditional tests pass again. I'm not aware of connection handling in Django ORM being different in test env...

selwin:
@patroqueeet I'm not a Celery user so I don't really know what `CELERY_ALWAYS_EAGER` does and can't really comment on this, nor do I have the time to learn how to set it up.

So for this one, I'll have to rely on your judgement on how to fix this and pull this in.

For further Celery related development issues, I'd require a proper CI with Celery setup before I'll accept any new contributions.

selwin:
Thinking more about this, are you able to change this integration to behave more like uWSGI? https://github.com/ui/django-post_office#integration-with-uwsgi

In this integration, it simply calls `call_command('send_queued_mail', processes=1)`.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/350

patroqueeet:
@selwin don't want to go for the uwsgi approach. all logic should be inside the django/celery env. sry.

utapyngo:
> one more: sry for bugging: the `connection.close()` kills the test suite because tests run with `CELERY_ALWAYS_EAGER`(there is no celery in CI) and then in this process db connection for all tests is closed. added a check for this locally:
> 
> ```
>  87                 # Close DB connection to avoid multiprocessing errors
>  88                 if not settings.CELERY_TASK_ALWAYS_EAGER:
>  89                     connection.close()
> ```

I faced the same issue when running tests.
How have you fixed it?

PR #350, opened 2020-11-25
Fix #326: queues emails twice

### LINKED ISSUE #326: Celery Task triggered by signal is not thread safe: Multiple E-Mails sent!

jrief:
This pull request shall fix #326 

I refactored the code from the management command into a utility function named `post_office.mail.send_all_in_queue()` – this name is open to discussion. That function then will also be called by the Celery task manager.

@selwin Something which is not yet clear to me, why do we have to close the database connection several times, while iterating through our loop?


post_office/mail.py
```diff
@@ -334,3 +337,25 @@ def send(email):
     )
 
     return len(sent_emails), num_failed, num_requeued
+
+
+def send_all_in_queue(lockfile=default_lockfile, processes=1, log_level=None):
```
selwin, line 342: What trips me up about this function name is the `all` part because it implies that everything that's queued will be sent, but in practice the ones that are scheduled for future delivery won't.

But I also can't think of a better name for this at this moment. I thought about `send_until_done()`, `enter_send_loop()` but these are also weird.

If we can't think of a better name, I think we should just rename this to `send_queued_mail` since this is also what the management command is called so there's consistency there at least.

post_office/management/commands/send_queued_mail.py
```diff
@@ -36,25 +28,5 @@ def add_arguments(self, parser):
...
-                    # Close DB connection to avoid multiprocessing errors
-                    connection.close()
-
-                    if not Email.objects.filter(status=STATUS.queued) \
-                            .filter(Q(scheduled_time__lte=now()) | Q(scheduled_time=None)).exists():
-                        break
-        except FileLocked:
-            logger.info('Failed to acquire lock, terminating now.')
\ No newline at end of file
+        logger.info('Acquiring lock for sending queued emails at %s.lock', options['lockfile'])
```
selwin, line 31: Can we move this logging call to inside `send_all_in_queue`? So this message will also appear when the lock is being acquired.

post_office/apps.py
```diff
@@ -10,5 +10,4 @@ def ready(self):
         from post_office import tasks
         from post_office.signals import email_queued
 
-        if hasattr(tasks, 'queued_mail_handler'):
-            email_queued.connect(tasks.queued_mail_handler)
+        email_queued.connect(tasks.queued_mail_handler)
```
selwin, line 13: I think this signal should only be connected if `CELERY_ENABLED` is `True`.

post_office/mail.py
jrief: I renamed that function to `send_queued_mail_until_done`. This reflects exactly what it does, and since that function is intended to be used internally, I don't have any problems with such a long name.
I strongly would discourage to reuse the name `send_queued_mail`. This can cause hefty confusion.

post_office/management/commands/send_queued_mail.py
jrief: done

post_office/apps.py
jrief: I'm not a friend of duck typing. The proposed solution uses a non-operational placeholder function [queued_mail_handler](https://github.com/ui/django-post_office/blob/78a15fa4d6b08cbeb0327c161d7d5533af8de682/post_office/tasks.py#L22) in case Celery is missing. This in my opinion is a more robust solution. Consider, if other queuing frameworks, such as for example [django-q](https://django-q.readthedocs.io/en/latest/) would be integrated into **Post-Office**.

jrief:
@selwin I also would strongly recommend to remove the `--lockfile` option from the management command. If a user chooses another lockfile, while Celery triggered `send_queued_mail`, we again might end sending emails twice again. Therefore, for this task there shall be one and only one hard-coded lockfile.

A lockfile might not even be enough. Consider the use case, when Post-Office is running on more than one instance. We maybe should rethink about adding something like a database mutex for this. This then would be the appropriate location for doing a `SELECT FOR UPDATE ...`. What do you think? Shall I implement a database mutex table and some locking mechanism?

selwin:
> A lockfile might not even be enough. Consider the use case, when Post-Office is running on more than one instance.

This is actually the intention of exposing `--lock-file` argument to the management command. User should be able to override this from cron or other places.

> If a user chooses another lockfile, while Celery triggered send_queued_mail, we again might end sending emails twice again. 

To solve this, I think we can add `PATH_TO_LOCKFILE` option to post office's config. But this should be done in a separate PR.

My first priority is to get this merged and release a bugfix version as soon as possible.

jrief:
OK, then if this PR is ready for merging, I can build a new version. Shall I?

jrief:
Ups, forgot to push my latest changes to GitHub.

post_office/apps.py
selwin: Checking whether `CELERY_ENABLED` settings is `True` is not duck typing.

If Celery is not enabled, there's no need to connect to this signal. If I remember correctly, Django has an optimization that if there's nothing connected to a signal, the signal won't be sent.

selwin:
I just realized that `CELERY_ENABLED` also defaults to `True`. This has to be changed to default to `False`. https://github.com/ui/django-post_office/blob/ac47aedbd30fbb80e63a5b312255cd81cc0fdc40/post_office/settings.py#L82

Otherwise having with celery installed will be automatically have this background task enabled without them knowing about it.

post_office/apps.py
jrief: > If Celery is not enabled, there's no need to connect to this signal. If I remember correctly, Django has an optimization that if there's nothing connected to a signal, the signal won't be sent.

This additional condition might safe a few milliseconds at the cost of more complexity. Anyway, I added this condition.

selwin:
Thanks!

Issue #218, opened 2017-12-20
SMTPServerDisconnected error with send_mail and Celery


mihalikv:
Hello, i am facing issue when i want to send_mail like this:
```
@task(name='send_registration_mail')
def send_registration_mail():
    # some stuff
    mail.send(
	    subject=template.subject,
	    message=message,
	    html_message=rendered_content,
	    priority='now'
	    **kwargs
	)
```
This code mostly ends failed with `SMTPServerDisconnected` in log '**please run connect() first**'. When i manualy requeue emails via admin and run `send_queued_mail`, emails are send.

dependencies:
```
django-post-office==3.0.3
Django==1.10.7

celery = 4.0.2
```

VerosK:
Hi, @mihalikv  I have similar issues on one machine of two deployed (both Celery 3.x, Django 1.10, PostOffice 3.0.4).  

Can you provide more details on your configration? I'm interested which Django mail backend and which MTA are you using.

yprez:
I'm getting the same problem... seems like an older issue that I missed.

It also happens in the shell, when leaving it open for long.

mihalikv:
@VerosK we use 
`EMAIL_POST_OFFICE_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'` and Microsoft Office.

Current set up is using `send_mail ` without celery task and created scheduled Celery task every minute like this:

```
@app.task(name='send_queued_mail')
def send_queued_mail():
    call_command('send_queued_mail')
```
with following settings:

```
POST_OFFICE = {
    'DEFAULT_PRIORITY': 'high'
}
```
we also have following celery task to send failed emails every 2 minutes:

```
@app.task(name='send_failed_mails')
def send_failed_mails():
    from post_office.models import Email, STATUS, Log

    max_retries = getattr(settings, 'EMAIL_MAX_RETRIES', 5)
    # select FAILED email WHERE num_logs <= max
    query_set = Email.objects.annotate(num_logs=Count('logs')).filter(Q(status__isnull=True) | Q(status=STATUS.failed)).filter(num_logs__lte=max_retries).all()
    logs = []
    for mail in query_set:
        if not mail.status:
            logger.error('Email without status with id: {} from: {}, to: {}'.format(mail.id, mail.from_email, mail.to))
        if mail.logs.count() == max_retries:
            logger.error('Unable to send mail with id: {} from: {}, to: {}'.format(mail.id, mail.from_email, mail.to))
[...]

PR #352, opened 2020-12-14
Fixes: #218 SMTPServerDisconnected error

weimens:
Reintroduce abandoned disconnect_after_delivery parameter for Email.dispatch.
The disconnect_after_delivery parameter got deleted in https://github.com/ui/django-post_office/commit/03d1a3771eaca0c226ca5a7e299f8752f3707c43.

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -183,6 +183,9 @@ class Email(models.Model):
             if not commit:
                 raise
 
+        if disconnect_after_delivery:
+            connections.close()
+
         if commit:
             self.status = status
             self.save(update_fields=['status'])


post_office/models.py
```diff
@@ -183,6 +183,9 @@ def dispatch(self, log_level=None,
             if not commit:
                 raise
 
+        if disconnect_after_delivery:
+            self.email_message().get_connection().close()
```
weimens, line 187:
```suggestion
            connections.close()
```
would also work, without failed tests, though it is not the exactly same behaviour as before. What is the preferred approach? Close all connection, rework the failed test or add an way to the ConnectionHandler to access cached connections without creating a new connection?

selwin:
Thanks!

PR #365, opened 2021-05-25
Improvement delete expired mails mechanism

IkhtiaryRizky:


post_office/utils.py
```diff
@@ -138,14 +138,22 @@ def parse_emails(emails):
...
     """
     Delete all emails before the given cutoff date.
     Optionally also delete pending attachments.
     Return the number of deleted emails and attachments.
     """
     expired_emails = Email.objects.only('id').filter(created__lt=cutoff_date)
-    emails_count, _ = expired_emails.delete()
+    total_deleted_emails = 0
+    while expired_emails:
+        email_ids = expired_emails.values_list('id', flat=True)[:batch_size]
```
ibadarrohman, line 150: dirubah jadi list, kalo nggak jadi nested query

post_office/utils.py
```diff
@@ -138,14 +138,22 @@ def parse_emails(emails):
...
+        
+        # Delete email and incr total_deleted_emails counter
+        _, deleted_data = Email.objects.filter(id__in=email_ids).only('id').delete()
```
ibadarrohman, line 153: kenapa pake `only` ya?

ga langsung ambil total_deleted_emails nya dari index yang pertama aja?
```PYTHON
deleted_emails, _ = Email.objects.filter(id__in=email_ids).delete()
total_deleted_emails += deleted_emails
```

IkhtiaryRizky: > why use `only`?, i thinks its not needed for delete operation
> 
> what about this?
> 
> ```python
> deleted_emails, _ = Email.objects.filter(id__in=email_ids).delete()
> total_deleted_emails += deleted_emails
> ```

Actually, I use "only" because referring to the previous code. But yeah, I think it's ok to be removed. For the `deleted_emails` variable, I think it's incorrect because the first element returned from the delete method is the total count for all deleted objects (including Log objects), so it's not an exact deleted email number. That's why I use an explicit deleted email number from the deleted_data['post_office.Email']. Please CMIIW.

post_office/utils.py
IkhtiaryRizky: > change to list first, i think this will make nested query

I think I can make it into a list of email ids start from the expired_emails variable, WDYT?

ibadarrohman: ah i think thats better, so no n-query executing `values_list` in looping

post_office/utils.py
ibadarrohman: i see, the logs will be deleted and counted too right

IkhtiaryRizky: Yeah, that's correct.

post_office/utils.py
```diff
@@ -138,14 +138,22 @@ def parse_emails(emails):
...
     """
     Delete all emails before the given cutoff date.
     Optionally also delete pending attachments.
     Return the number of deleted emails and attachments.
     """
-    expired_emails = Email.objects.only('id').filter(created__lt=cutoff_date)
-    emails_count, _ = expired_emails.delete()
+    expired_emails = Email.objects.only('id').filter(created__lt=cutoff_date).values_list('id', flat=True)
+    total_deleted_emails = 0
+    while expired_emails:
```
ibadarrohman, line 149: this will evaluate the queryset to hit DB to check expired email is existing every time loop continue, and on line 150 will hit DB again to limit it based on batch_size

what i mean is before the loop we need to fetch expired email ids first
then batch the ids to delete then delete it
```
expired_email_ids = list(Email.objects.only('id').filter(created__lt=cutoff_date).values_list('id', flat=True))
total_expired_emails = len(expired_email_ids)
offset = 0
email_ids = expired_email_ids[offset:batch_size]
while email_ids:
   deleted_data = Email.objects.filter(id__in=email__ids).delete()
   offset += deleted_data['post_office.Email']
   email_ids = expired_email_ids[offset:offset + batch_size]
```

jrief:
Just a question. Why do you add `batch_size`, and if the number of emails to be deleted exceeds 10000, repeat the loop?Wouldn't it be better to then just stop and wait until Celery runs that task again? 

Moreover, if you introduce `batch_size`, then I would expect that parameter to be passed in through the management command.

post_office/utils.py
```diff
@@ -138,14 +138,22 @@ def parse_emails(emails):
     return emails
 
 
-def cleanup_expired_mails(cutoff_date, delete_attachments=True):
+def cleanup_expired_mails(cutoff_date, delete_attachments=True, batch_size=10000):
```
selwin, line 141: Thinking about this, can you make this lower to 1k?

@jrief is also right, do you mind adding a `--batch-size` argument to the management command also?

selwin:
> Wouldn't it be better to then just stop and wait until Celery runs that task again?

Some people only runs this task daily via cron so I think it makes sense. I also ran into this issue in my project where after a few years of operation unable to delete the emails so I had to manually clean them up.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/386

PR #379, opened 2021-09-29
add re-dispatch functionality in the admin

domdinicola:
It would be nice to have the utility to resend an email through the admin

### DIFF
--- a/post_office/admin.py
+++ b/post_office/admin.py
@@ -1,14 +1,15 @@
 import re
 
 from django import forms
-from django.db import models
-from django.contrib import admin
 from django.conf import settings
+from django.contrib import admin, messages
 from django.core.exceptions import ValidationError
 from django.core.mail.message import SafeMIMEText
+from django.db import models
 from django.forms import BaseInlineFormSet
 from django.forms.widgets import TextInput
-from django.http.response import HttpResponse, HttpResponseNotFound
+from django.http.response import (HttpResponse, HttpResponseNotFound,
+                                  HttpResponseRedirect)
 from django.template import Context, Template
 from django.urls import re_path, reverse
 from django.utils.html import format_html
@@ -16,13 +17,15 @@ from django.utils.text import Truncator
 from django.utils.translation import gettext_lazy as _
 
 from .fields import CommaSeparatedEmailField
-from .models import Attachment, Log, Email, EmailTemplate, STATUS
+from .mail import send
+from .models import STATUS, Attachment, Email, EmailTemplate, Log
 from .sanitizer import clean_html
 
 
 def get_message_preview(instance):
     return ('{0}...'.format(instance.message[:25]) if len(instance.message) > 25
             else instance.message)
+
 
 get_message_preview.short_description = 'Message'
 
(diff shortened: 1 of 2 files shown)


post_office/admin.py
```diff
@@ -201,6 +206,19 @@ def fetch_email_image(self, request, pk, content_id):
                 return HttpResponse(message.get_payload(decode=True), content_type=message.get_content_type())
         return HttpResponseNotFound()
 
+    def resend(self, request, pk):
```
selwin, line 209: This is definitely a feature that's nice to have. However I'd like to change the behavior a little bit.

Instead of creating a new email, it should just fetch the current email and do an `email.dispatch()`, essentially just doing a resend on the email.

domdinicola: absolutely!

post_office/admin.py
```diff
@@ -201,6 +206,12 @@ def fetch_email_image(self, request, pk, content_id):
...
```
selwin, line 209: Another request, can you make this a POST request?

post_office/templates/admin/post_office/email/change_form.html
```diff
@@ -0,0 +1,5 @@
+{% extends "admin/change_form.html" %}{% load i18n %}
+{% block object-tools-items %}
+    {{ block.super }}
+    <li><a href="{% url 'admin:resend' original.pk %}" style="background-color:#0099ff">{% trans "Resend" %}</a></li>
```
domdinicola, line 4: @selwin to make it a POST, I would need to make a form here or you do have another idea

domdinicola: <form method="post" url=reverse url>
  <button type="submit">
  </button>
</form>

selwin: Yes, we should make a django admin button. It should be something like this https://stackoverflow.com/a/45072862

domdinicola: it's still a get since is using <a href..
it's already a button tho

domdinicola: Screen Shot 2021-12-12 at 4 34 38 PM
this is the look and feel

domdinicola: @selwin do you think this is ok?

selwin: I'm ok with the look and feel. If you can make it a POST request, it would be great because it's more technically correct.

selwin: > to make it a POST, I would need to make a form here or you do have another idea

Yes, I think this is the right solution solution.

selwin:
@domdinicola do you mind resolving the conflicts and making this a POST request so I can merge this in?

domdinicola:
@selwin conflicts have been merged.
If you have advices to how proceed with a POST please help.
The link you provided (https://stackoverflow.com/a/45072862) is using a GET too and django_object_actions too.

domdinicola:
@selwin are we good here?

selwin:
Thanks!

Issue #325, opened 2020-07-31
Rendered preview in admin


WoLpH:
In my fork of the project I have a rendered preview of the email which is a really useful feature if you ask me. It shows both the text and html variant of the mail in an iframe within the admin.

Would this be an interesting addition for the project?

The changes needed:
1. A `example_context` field added to the `EmailTemplate` model
2. A few admin changes: https://github.com/ui/django-post_office/compare/master...WoLpH:master#diff-43bb3346d5574d3cfa6043e724bcc7f7

I can work on making a clean pull request if this is something that would be merged :)

jrief:
You mean something [like this](https://github.com/ui/django-post_office/pull/314)?

This will be available in the next major release, presumable 3.5

WoLpH:
Yes. That's very similar 👍

WoLpH:
@jrief I'm reopening this because that feature is similar but not nearly as useful as the one I'm proposing here :)

It currently only shows the rendered version of sent emails. My code shows how an email template will look for both HTML and text before you send it using example context passed along as JSON data. Additionally, it also shows the variants in all other languages if available.

jrief:
I see! So basically you show how future emails will be get rendered, while my former pull request show how sent emails have been rendered. Yes, I believe inside the post_office template view, this could be quite useful. Do you have a screenshot of that feature?

WoLpH:
This is how it currently looks. The only data model change it requires (I think) is the addition of a json field to specify the template variables.

I'm using iframes with separate requests for the rendering. The advantage of this method is that if you have a template syntax error or something it won't break the admin page

![image](https://user-images.githubusercontent.com/270571/99208223-31f0c980-27c0-11eb-8d40-0c5d25c12a43.png)

WoLpH:
You can see the current diff here: https://github.com/ui/django-post_office/compare/master...WoLpH:master

jrief:
This feature would really be useful for drafting email templates.

WoLpH:
Should I prepare a clean pull request in that case?
Any changes you would like in the workings?

PR #383, opened 2021-11-30
Add rendered preview in admin. Fixes: #325

WoLpH:


post_office/admin.py
```diff
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
jrief, line 2: please remove the two lines above. In python-3 they don't make any sense.

post_office/admin.py
```diff
@@ -9,6 +12,7 @@
 from django.core.mail.message import SafeMIMEText
 from django.forms import BaseInlineFormSet
 from django.forms.widgets import TextInput
+from django.utils import html
```
jrief, line 15: check order of import statements

post_office/admin.py
```diff
@@ -256,11 +262,33 @@ class EmailTemplateInline(admin.StackedInline):
...
+              'rendered_content', 'rendered_html_content',)
+    readonly_fields = 'rendered_content', 'rendered_html_content'
     formfield_overrides = {
         models.CharField: {'widget': SubjectField}
     }
 
+    def rendered_content(self, instance):
+        if instance.content:
+            height = instance.content.count('\n') * 25
+            return html.mark_safe(
```
jrief, line 275: use `format_html`

post_office/admin.py
```diff
@@ -256,11 +262,33 @@ class EmailTemplateInline(admin.StackedInline):
...
+                '<iframe '
+                'style="width: 80%; height: {}px;"'
+                ' src="?preview=text&language={}">'
+                '</iframe>'.format(height, instance.language))
+        else:
+            return ''
+
+    def rendered_html_content(self, instance):
+        if instance.html_content:
+            return html.mark_safe('<iframe '
```
jrief, line 285: detto

post_office/admin.py
```diff
@@ -270,18 +298,69 @@ class EmailTemplateAdmin(admin.ModelAdmin):
     form = EmailTemplateAdminForm
     list_display = ('name', 'description_shortened', 'subject', 'languages_compact', 'created')
     search_fields = ('name', 'description', 'subject')
+    readonly_fields = 'rendered_content', 'rendered_html_content'
```
jrief, line 301: no implicit tuples

post_office/admin.py
```diff
@@ -270,18 +298,69 @@ class EmailTemplateAdmin(admin.ModelAdmin):
...
+                    template_instance.html_content
+                    .replace('inline_image', 'static')
+                    .replace(' post_office ', ' static '))
+            else:
+                template = engine.from_string(
+                    '<pre>%s</pre>' % template_instance.content)
+
+            return http.HttpResponse(template.render(instance.example_context))
+
+        return super(EmailTemplateAdmin, self).change_view(
```
jrief, line 343: this is Python-2 `super()`-style.

post_office/admin.py
```diff
@@ -270,18 +298,69 @@ class EmailTemplateAdmin(admin.ModelAdmin):
...
+            request, object_id, form_url=form_url, extra_context=extra_context)
+
+    def rendered_content(self, instance):
+        if instance.content:
+            height = instance.content.count('\n') * 25
+            return html.mark_safe(
```
jrief, line 349: detto

post_office/admin.py
```diff
@@ -270,18 +298,69 @@ class EmailTemplateAdmin(admin.ModelAdmin):
...
+                '<iframe '
+                'style="width: 80%; height: {}px;"'
+                ' src="?preview=text">'
+                '</iframe>'.format(height))
+        else:
+            return ''
+
+    def rendered_html_content(self, instance):
+        if instance.html_content:
+            return html.mark_safe('<iframe '
```
jrief, line 359: detto

jrief (changes requested):
Before fixing my code annotations, I'd rather rethink, if it doesn't make more sense to se a WYSIWYG-Editor straight ahead. This would be my preferred long term solution.

Thanks for your work, but this new feature must be approved by @selwin anyway. As I said, this shall only be a temporary solution.

post_office/admin.py
WoLpH: I'm not sure where those are coming from to be honest. I've never added them so it might be related to git merging things

WoLpH:
The pull request definitely needs cleanup, but since I didn't get a reply in the original issue anymore I thought it would be good to create a pull request so it doesn't get lost :)

Personally I think that this addition is useful even if you would have a WYSIWYG editor. The renderings of those editors are almost always different from a regular render. It's not my call however, if the WYSIWYG editor is the preferred solution than I'll close the pull request and wait for that release.

With that in mind... a form to send out a test-mail would be very useful as well.

selwin:
Having rendered HTML preview in admin would be great.

Aside from @jrief 's comments, please make sure that all JS are stripped to prevent XSS attacks.

WoLpH:
> Having rendered HTML preview in admin would be great.
> 
> Aside from @jrief 's comments, please make sure that all JS are stripped to prevent XSS attacks.

The easiest and safest way would be to sandbox the iframe, is that enough or should I also do a few search/replaces?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox

selwin:
@WoLpH we have https://bleach.readthedocs.io/ as optional dependency. We can use that to easily strip all JS.

selwin:
I think we can even make `bleach` a required dependency @jrief . This rendered preview is a big quality of life improvement in my opinion.

WoLpH:
I've cleaned up the commits and added the `clean_html()` call.

I should note that there is another similar attack vector in the code-base outside of this addition:
https://github.com/ui/django-post_office/blob/5297f00d886406de1233de6697262096e9ea81ca/post_office/admin.py#L179-L182

And possibly this one:
https://github.com/ui/django-post_office/blob/5297f00d886406de1233de6697262096e9ea81ca/post_office/admin.py#L173-L177

post_office/models.py
```diff
@@ -269,6 +269,7 @@ class EmailTemplate(models.Model):
         default='', blank=True)
     default_template = models.ForeignKey('self', related_name='translated_templates',
         null=True, default=None, verbose_name=_('Default template'), on_delete=models.CASCADE)
+    example_context = context_field_class(_('Context'), blank=True, null=True)
```
selwin, line 272: Why are we adding a new field to the model? Adding a preview to admin shouldn't necessitate adding a new field.

post_office/admin.py
```diff
@@ -249,36 +256,111 @@ def __init__(self, *args, **kwargs):
...
+                       'rendered_html_content'),
+        }),
     ]
     inlines = (EmailTemplateInline,) if settings.USE_I18N else ()
     formfield_overrides = {
         models.CharField: {'widget': SubjectField}
     }
+    change_form_template = 'admin/post_office/EmailTemplate/change_form.html'
+
+    def change_view(self, request, object_id, form_url='', extra_context=None):
```
selwin, line 325: Can you move this dedicated preview functionality to a new view?

post_office/models.py
WoLpH: I think the variable name is self explanatory, but it's to add a context to the rendered preview so all variables can be rendered with a useful value.

selwin:
@WoLpH gentle reminder to update this PR so we can merge this in.

gabn88:
@WoLpH @selwin This sounds like a great feature that I might use right away in my project. I have been reading the code and the idea from the beginning. And it makes me ask myself, should we need an extra STATUS_CHOICES on the Email Model? Something like 'draft'? Or am I missing something?

PR #392, opened 2022-02-02
Add missing migration file. Fixes: #391

### LINKED ISSUE #391: Running `makemigrations --check --dry-run` fails because of missing migration file.
Running the command `django-admin makemigrations --check --dry-run` results with exit code `1`, because the project is missing a migration file.

CirXe0N:
Closes #391


selwin:
Hmm not sure whether this is neeed. Is this caused by Django defaulting to use big integer field as primary key?

perepicornell:
Exactly.
Why would it not be needed?

El dc., 25 de maig 2022, 14:38, Selwin Ong ***@***.***> va
escriure:

alexandernst:
No, this is not required. The entire problem comes from [this Django change](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys)

For example, celery beat fixed it by adding `default_auto_field = 'django.db.models.AutoField'` to it's `apps.py` ([MR](https://github.com/celery/django-celery-beat/pull/400/files)).

If you could add this, no migrations would be required. /cc @selwin

CirXe0N:
The solution that @alexandernst suggested seems like a better implementation. @selwin If I make a MR with that change, would that get merged?

CirXe0N:
There is a new [MR](https://github.com/ui/django-post_office/pull/401) with the solution by setting the default auto field. Therefore I am closing this MR.

PR #394, opened 2022-03-02
[Proposal] - Django builtin JSONField + metadata Email field

slim0:
Hi,

It is just a proposal. But since the changes are minor, I created a pull request.

At the beginning, I was just looking for a custom metadata field. It can be useful to tag and retrieve emails.

Then I realize that you don't use the builtin Django JSONField and I was wondering why ?

migration file was added.

Please let me now.

Thx.


jrief:
In general, you should not mix two independent changes.

Please create a pull request for changing the external `JSONField` to Django's internal one added in version 3.1. Check that the allowed versions are up-to-date – [they aren't](https://github.com/ui/django-post_office/blob/2f78c5ea5d3828362e9bcc723de1176146f02e74/setup.py#L50) and add that change to the changelog.

Then as next step provide another pull request to add the `metadata` field. Please provide some insight, why this can be helpful for the project.

PR #400, opened 2022-04-19
Fix: use bleach 5.0.0 css sanitizer

franciscobmacedo:
bleach 5.0.0 `clean` method should include `css_sanitizer`.


fasih:
I think we need to add additional dependencies `bleach[css]` in the `setup.py` file as well

`install_requires=['django>=2.2', 'jsonfield>=3.0', 'bleach', 'pytz', 'bleach[css]']`

franciscobmacedo:
> Is this PR backward compatible with bleach <5.0? @franciscobmacedo

I guess it doesn't, should i require `bleach>=5.0` in the `install_requires`?

selwin:
Can you slightly modify this PR to fallback to the old behavior on bleach < 5? I prefer to not pin dependencies when I can. If we force people to upgrade, this may also affect code in their own projects.

franciscobmacedo:
sure, will do

selwin:
@franciscobmacedo do you mind resolving the conflicts on `setup.py`?

franciscobmacedo:
I'm sorry for the delay, I haven't been that active here recently.

selwin:
Thanks!

domdinicola:
is this going to be released soon?

franciscobmacedo:
I think it is released in [3.6.1](https://github.com/ui/django-post_office/releases/tag/v3.6.1)

domdinicola:
I think the package does not include the templates...

PR #403, opened 2022-07-13
Fix: save template even if not rendered on delivery (fix #395)

### LINKED ISSUE #395: Value of USE TEMPLATE in admin is wrong, but template is used
I created a email template in database admin. And the sent mail with this command:

            mail.send(
                'recipient@example.com', # List of email addresses also accepted
                'from@example.com',
                template='user_cancelled_subscription', # Could be an EmailTemplate instance or name
                context={'foo': 'bar'},
            )

Email subject and content is filled with values from the template, everything seems to work prefectly. But it admin, there is a red cross below USE TEMPLATE:
image

And when I click filter `user_cancelled_subscription` for the "By Name" filter, the email is not shown.

What could be wrong? This is really confusing.

franciscobmacedo:
When creating an email in the `create` function (`mail.py`) without the `render_on_delivery` option, it should still save the template, as it was still used and can be usefull for admin purposes.

This should solve #395

### DIFF
--- a/post_office/mail.py
+++ b/post_office/mail.py
@@ -86,7 +86,8 @@ def create(sender, recipients=None, cc=None, bcc=None, subject='', message='',
             expires_at=expires_at,
             message_id=message_id,
             headers=headers, priority=priority, status=status,
-            backend_alias=backend
+            backend_alias=backend,
+            template=template,
         )
 
     if commit:


selwin:
Thanks!

JiriKr:
Thanks for this change :+1: . Maybe we should modify also  https://github.com/ui/django-post_office/blob/dddde78411a28f8a5264cfd5244661c6cb75ab88/post_office/models.py#L108 ?

PR #407, opened 2022-07-22
Added the draft status to the Email model

gabn88:
This is to accomodate for previewing an email before it is send (or queued).

It is quite straightforward, but in case there are any questions please let me know.


selwin:
I think this feature is useful. But I'm not sure that this status should be added to `email.status` (as this relates to delivery status). I was thinking it should be a separate field like `publishing_status` or something like that.

But if this is made into an additional field, I do wonder whether we can do anything more with it.

Issue #406, opened 2022-07-19
HTML Email sending issue with with django-allauth custom templates


fuser-henry:
I recently upgraded from django_post_office version 3.4.0 to 3.6.1 and some of my email templates have broken. 
The emails would send, but the HTML wouldn't be rendered properly (subject line is all correct). 

I have no problem with templates / html templates that are put into django_post_office, however, with html templates used by django-allauth, they would work before the update but not afterwards. 

It seems that this may be some changes made from 3.4.0 to 3.6.1, more specifically, regarding adding the dependency for bleach, which may have changed the behavior in regards to templates. 
I have moved back and forth between version 3.4.0 and 3.6.1 as the only change to reproduce the problem reliably. Uninstalling the optional bleach dependency didn't resolve my problem. 

Would anyone have any suggestions / ideas of what might have caused this?

Gutei:
You can try  to added parameter message when calling `send` @fuser-henry , it's work for me
```Python
mail.send(
    ...
    message=strip_tags(html_message),
    ...
)
````

gabn88:
I have found the issue and described it in detail here: https://github.com/ui/django-post_office/issues/419

Please close this issue and also the pull request, as it is not fixing the underlying issue.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/420

PR #411, opened 2022-08-16
fix HTML wouldn't be rendered in email body

Gutei:
resolve issue [406](https://github.com/ui/django-post_office/issues/406).
added calling `attach_alternative()` method for including extra versions of the message body in the email


selwin:
@Gutei mind creating a test case for this?

PR #416, opened 2022-09-19
Fix performance of EmailAdmin view.

### LINKED ISSUE #415: Poor performance in EmailAdmin view.
There is a performance issue when viewing an Email in the EmailAdmin when there are a lot Emails and Attachments in the database. This is caused by customizing the `get_queryset` method of `AttachmentInline` to try to filter out inline attachments by their header values.

In the `post_office.admin.AttachmentInline.get_queryset` method, Django has not yet applied the Email FK filter to the queryset. That happens later when the inline formset is being created.

That means that the query being run in the `get_queryset` method actually fetches and iterates over every single Email Attachment object in the database, regardless of the Email it is for. In the loop, it also references the `attachment` property, but without a `select_related` on the query, it will have to fetch each attachment separately for every iteration.

If you have a lot of Emails and Attachments in the database, this causes performance issues.

barqshasbite:
Add a `filter` and `select_related` to the queryset to limit the records being fetched by the custom `get_queryset` logic.

Fixes #415

### DIFF
--- a/post_office/admin.py
+++ b/post_office/admin.py
@@ -40,9 +40,19 @@ class AttachmentInline(admin.StackedInline):
         are displayed anyway.
         """
         queryset = super().get_queryset(request)
+
+        # The queryset does not have the Email FK filter applied at this point.
+        # That happens later when the inline formset is created as part of the
+        # admin view. Here we explicitly add the Email filter to avoid selecting
+        # every row in the table during this special filtering step.
+        email = request.resolver_match.kwargs.get("object_id", None)
+        if email is None:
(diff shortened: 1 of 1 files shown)


selwin:
@barqshasbite mind fixing the conflict so I can merge this in? Hoping to cut a release this weekend or next week at the latest.

petrprikryl:
It has been fixed in #413. And `select_related` is not more needed due #414.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/413

Issue #419, opened 2022-10-06
HTML emails not stored correctly in database


gabn88:
Django is (sometimes) encoding the html part of EmailMessage as quoted_printable. In the following line we get this html message:
https://github.com/ui/django-post_office/blob/13aaac0d1575fb4805d71bc7f2ee390741eaec41/post_office/backends.py#L45

It might look like (message.as_string):
[...]

PR #420, opened 2022-10-13
Fix saving html emails that have quoted_printable transfer

gabn88:
See https://github.com/ui/django-post_office/issues/419

### DIFF
--- a/post_office/backends.py
+++ b/post_office/backends.py
@@ -33,7 +33,9 @@ class EmailBackend(BaseEmailBackend):
                 headers.setdefault("Reply-To", reply_to_header)
             message = email_message.body # The plaintext message is called body
             if hasattr(email_message, 'alternatives') and len(email_message.alternatives) > 0:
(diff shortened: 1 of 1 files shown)


selwin:
Mind adding a quick test case to this one? Just want to make sure that the result is correct.

gabn88:
@selwin I have skipped the whole decoding part. This should fix the problem without any decoding. As far as I can tell this is ready to merge.

gabn88:
@selwin I can think of two tests that are missing:
1. Sending email without html
2. Sending email with quoted-printable Transfer type

The second one is hard to test, as I don't know when the python EmailMessage changes the encoding type. However, I have now tested this code thoroughly on production sites. And it fixes a major bug (email not being sent correctly, html emails as well as plain/text emails).

gabn88:
@selwin please approve this PR as it solves a critical flaw in django-post_office.

selwin:
Thanks!

PR #423, opened 2022-10-18
Fix: templates are not included in package (closes #404)

### LINKED ISSUE #404: templates are not included in the package

zagl:
Includes the template HTML files in MANIFEST.in so that they get included into the pip package.
closes #404


domdinicola (approved):
exactly what we need!

PR #459, opened 2023-12-27
Fix #458

### LINKED ISSUE #458: Multiple instances on the same server
Hi, I have multiple instances of my application deployed on one server. I got the error, stacktrace attached
```
FileExistsError: [Errno 17] File exists: '/tmp/916412.lock' -> '/tmp/post_office.lock'
  File "celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
  File "post_office/tasks.py", line 33, in send_queued_mail
    send_queued_mail_until_done()
  File "post_office/mail.py", line 470, in send_queued_mail_until_done
    with FileLock(lockfile):
  File "post_office/lockfile.py", line 148, in __enter__
    self.acquire()
  File "post_office/lockfile.py", line 128, in acquire
    os.symlink(self.pid_filename, self.lock_filename)
```

I use celery option.
It looks like I cannot fix that problem without monkey-patching this
https://github.com/ui/django-post_office/blob/9acc56f2c38f36862c296fea587879dc776f80a8/post_office/lockfile.py#L155
or rewriting tasks https://github.com/ui/django-post_office/blob/9acc56f2c38f36862c296fea587879dc776f80a8/post_office/tasks.py

Ideally name of lock file should be a setting like
```python
POST_OFFICE = {
    "BACKENDS": {
        "default":  ...
    },
    "DEFAULT_PRIORITY": "now",
    "CELERY_ENABLED": True,
    "LOCK_FILE_NAME": "post_office_...",
}
```

```python
 default_lockfile = os.path.join(tempfile.gettempdir(), settings.LOCK_FILE_NAME) 
```

Mogost:
Fix for #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,17 @@ POST_OFFICE = {
 }
 ```
 
+### Lock File Name
+The default lock file name is 'post_office', but this can be altered by setting `LOCK_FILE_NAME` in the configuration.
+
+```python
+# Put this in settings.py
+POST_OFFICE = {
+    ...
+    'LOCK_FILE_NAME': 'custom_lock_file',
+}
+```
+
 ### Override Recipients
 
 Defaults to `None`. This option is useful if you want to redirect all
--- a/README.md
+++ b/README.md
@@ -453,7 +453,7 @@ POST_OFFICE = {
 ```
 
 ### Lock File Name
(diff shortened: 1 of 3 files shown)


selwin:
This is a very useful addition. Do you mind documenting this?

Mogost:
Done

selwin:
Thanks!

PR #467, opened 2024-05-14
Optimize AttachmentAdmin

petrprikryl:
It fixes problem of long rendering time at attachment detail admin page when you have a lot entries in DB. It takes few minutes in our environment because of 170k options in select.

image


selwin:
Thanks!

Issue #421, opened 2022-10-18
Already rendered Email is re-rendered without context


productaize:
In closing #403 a change was made to save the template independent of the render_on_delivery flag. This now causes the email to be re-rendered, however without the context. The result is the empty template message (placeholders are replaced with blank).

Either this should be changed to only re-render if Email.message or Email.html_message is empty, or the context should always be stored, as hinted here [on 24 Aug](https://github.com/ui/django-post_office/pull/403#issuecomment-1226085062)

[django-post_office/post_office/models.py](https://github.com/ui/django-post_office/blob/dddde78411a28f8a5264cfd5244661c6cb75ab88/post_office/models.py#L108)

zagl:
If the context is also just stored and nothing else changed it would be `render_on_delivery=True` by default.

If the context also "can be usefull for admin purposes" it should be saved and Email.message and Email.html_message should be checked for beeing empty.

If the context should not be saved the line
https://github.com/ui/django-post_office/blob/dddde78411a28f8a5264cfd5244661c6cb75ab88/post_office/models.py#L108
could simply be replaced by
`if self.context is not None:`

I created a pull request for the latter.

gabn88:
This commit creates a problem with emails that have a context, but no template. I get "AttributeError: 'NoneType' object has no attribute 'subject'". 

`subject = engine.from_string(self.template.subject).render(self.context)`

PR #468, opened 2024-06-02
Successfully prepare email with context but without template

pacahon:
This change should fix the case when Email instance stores context, but not the template.

It's quite useful to store context for debug purposes even if you store already rendered message.

https://github.com/ui/django-post_office/issues/421#issuecomment-1282609243

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -104,7 +104,7 @@ class Email(models.Model):
         if get_override_recipients():
             self.to = get_override_recipients()
 
-        if self.context is not None:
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

pacahon:
@selwin Hi! Is it possible to release this fix on pypi in the near future?

selwin:
Done, 3.9.0 is now up on pypi. https://pypi.org/project/django-post-office/3.9.0/

PR #470, opened 2024-06-12
Upgrade python>=3.9

Mogost:
3.8 is near to eol
@selwin could you check this please

### DIFF
--- a/post_office/lockfile.py
+++ b/post_office/lockfile.py
@@ -44,7 +44,7 @@ class FileLock:
     def get_lock_pid(self):
         try:
             return int(open(self.lock_filename).read())
-        except IOError:
+        except OSError:
             # If we can't read symbolic link, there are two possibilities:
             # 1. The symbolic link is dead (point to non existing file)
             # 2. Symbolic link is not there
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,7 @@ name: Test
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
 
 permissions:
   contents: read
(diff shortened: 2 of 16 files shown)


post_office/lockfile.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, lock_filename, timeout=None, force=False):
     def get_lock_pid(self):
         try:
             return int(open(self.lock_filename).read())
-        except IOError:
+        except OSError:
```
selwin, line 47: Any reason why this is changed?

selwin:
Thanks, this is a helpful update.

Ideally we should also remove `bleach` dependency as it's no longer maintained.

post_office/lockfile.py
Mogost: In Python 3.3+ this exceptions are now simple alias. OSError is actually thrown

Mogost:
> Thanks, this is a helpful update.
> 
> Ideally we should also remove `bleach` dependency as it's no longer maintained.

I am currently in the process of refining and enhancing the project. Following the completion of this pull request, I intend to integrate additional linters and code formatters. By dividing the work into manageable segments, I aim to facilitate more efficient and less burdensome code reviews.

selwin:
> I am currently in the process of refining and enhancing the project. Following the completion of this pull request, I intend to integrate additional linters and code formatters. By dividing the work into manageable segments, I aim to facilitate more efficient and less burdensome code reviews.

Sounds like a plan. Will merge this in.

selwin:
If you get around to adding code formatters (I'm assuming `black` or `ruff --format`), please use 120 characters as the max line length and `--skip-string-normalization` so it doesn't automatically change single quote to double quotes.

Mogost:
> If you get around to adding code formatters (I'm assuming `black` or `ruff --format`), please use 120 characters as the max line length and `--skip-string-normalization` so it doesn't automatically change single quote to double quotes.

Ruff is my choice. 120 OK, no problem. But I don't want to skip string normalization. It's good for consistency. I can split skip normalization in separated pr but in general I will be happy to see it enabled.

selwin:
> But I don't want to skip string normalization. It's good for consistency. I can split skip normalization in separated pr but in general I will be happy to see it enabled.

Sorry, I happen to belong in the camp that only started using black after `--skip-string-normalization` was introduced :). I think this issue has been debated to death in https://github.com/psf/black/issues/118 and in the end it just comes down to preference.

Mogost:
> > But I don't want to skip string normalization. It's good for consistency. I can split skip normalization in separated pr but in general I will be happy to see it enabled.
> 
> Sorry, I happen to belong in the camp that only started using black after `--skip-string-normalization` was introduced :). I think this issue has been debated to death in [psf/black#118](https://github.com/psf/black/issues/118) and in the end it just comes down to preference.

We can normalize strings to `'` instead of `"`. I am speaking only about consistency. It does not matter to me if it is consistent `'` or consistent `"`

selwin:
> We can normalize strings to ' instead of ". I am speaking only about consistency. It does not matter to me if it is consistent ' or consistent "

Ok.

PR #474, opened 2024-06-19
Add isort

Mogost:
@selwin I have applied only isort rules for simplifying the review process.

Also, I have enforced import per line as it is a more merge-friendly format.


Mogost:
@selwin If you want another configuration let me know

selwin:
@Mogost sorry I just saw this. Mind bumping the ruff version used in the workflow file? Also, I'm not a fan of using single line imports, ming removing this?

Mogost:
Hi @selwin,

Thanks for your feedback! I prefer to stick with single-line imports as they significantly reduce merge conflicts, especially when multiple contributors are working on the same file. They also make diffs clearer by isolating changes to individual imports, which simplifies reviews. 

This format does increase the length of import blocks, making source code files longer.
But I find this is a small price to pay for the reduced merge conflicts. And the cost is limited by using editor features, such as “go to definition” for navigation, or PyCharm’s default to “folding” of import blocks. Even if it only prevents a few conflicts per month, I think it’s worth it.

selwin:
> Thanks for your feedback! I prefer to stick with single-line imports as they significantly reduce merge conflicts, especially when multiple contributors are working on the same file. They also make diffs clearer by isolating changes to individual imports, which simplifies reviews.

I understand your argument and I do acknowledge that merge conflicts will also be reduced. I'm just not ready to move on from my decades old habit of how I see modules are imported in Python 🙏🏻

PR #480, opened 2025-04-30
Add post office email object reference to EmailMessage

chromium7:
I need to implement  a custom `BaseEmailBackend` and override its `send_messages` function which takes a list of django `EmailMessage`. After sending the message, I need to save the external email ID to the db. By providing a reference to the email object, we wont  need to query for the email again and we can update easier.

### DIFF
--- a/post_office/messages.py
+++ b/post_office/messages.py
@@ -1,6 +1,9 @@
+from typing import TYPE_CHECKING
+
 from django.core.mail import EmailMessage, EmailMultiAlternatives
 
-from .models import Email
+if TYPE_CHECKING:
+    from .models import Email
 
 
 class PostOfficeEmailMessage(EmailMessage):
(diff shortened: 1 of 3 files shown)


bakumpai (approved):

ibadarrohman (commented):
Methods (prepare_email_message and email_message)  now returns PostOfficeEmailMessage or PostOfficeEmailMultiAlternatives, which are subclasses of the original Django classes. While this maintains compatibility at the subclass level, it could potentially break external code that relies on strict type checks (type() instead of isinstance()) or specific behaviors of the original classes.

My concern is, as this is a public library, such a change might have a global impact on users relying on the current behavior.

chromium7:
closing this PR since we found an alternative solution to use internal message ID instead of the external ID

PR #481, opened 2025-05-08
Better string representations for Email model

blag:
I tweaked the way the email model is stringified so it displays better in the admin. But I kept the internal `__repr__` method the way it was previously.

Note the breadcrumbs and the subtitle.

Previous admin:

previous django admin that displays email "to" addresses in a Python-esque syntax list

New admin:

new django admin that displays email "to" addresses in a comma-separated format

As a bonus, I also added Django 5.2 to the GitHub Actions test matrix.


selwin:
Thanks!

PR #482, opened 2025-05-13
Convert setup.py to pyproject.toml

blag:
I tried to do the bare minimum to convert from the deprecated `setup.py` to the new `pyproject.toml` declarative config.


pyproject.toml
```diff
@@ -1,3 +1,62 @@
...
+[project.optional-dependencies]
+test = [
+	"tox >= 2.3",
+]
+prevent-xss = [
+	"bleach",
+]
+testing = [
+	"tox >= 2.3",
+]
```
selwin, line 50: The indentation on these lines look a little bit too deep. Mind changing the indentation to 4 spaces and keep everything consistent?

blag: My tools apparently switched to using tabs halfway through. Good catch! Fixed.

selwin:
Thanks!

PR #483, opened 2025-05-13
Add post_office.contrib.sendgrid app

blag:
Just tossing this out there. If there's positive interest I'll include some tests and documentation.

Add a contrib app to include a Sendgrid Events webhook handler for deliverability events. This allows django-post_office to store emails and deliverability logs in the database -- which is exposed through the admin -- instead of relying on Sendgrid's deliverability logs.

Builds on top of #482 and sklarsa/django-sendgrid-v5#138.


selwin:
I like the idea of adding ESP notification/webhook handling. I myself use both SES and Sparkpost for work so I can contribute them to this repo.

Our task would be to design a coherent set of APIs that allow users to use these in their own projects. I think the views should be class based so users can implement their own behaviors on how they'd handle notification events.

blag:
I thought about making it a CBV, but I ultimately decided against it. CBVs allow developers to define a method that matches an HTTP verb (`get`, `post`, etc.), and then the CBV routes matching requests to that method.

CBVs are initialized in a URLConf with [their `CBV().as_view()` method](https://docs.djangoproject.com/en/5.1/ref/class-based-views/base/#django.views.generic.base.View.as_view). When a request is received, it is routed to the `setup` method:

> When the view is called during the request/response cycle, the [setup()](https://docs.djangoproject.com/en/5.1/ref/class-based-views/base/#django.views.generic.base.View.setup) method assigns the [HttpRequest](https://docs.djangoproject.com/en/5.1/ref/request-response/#django.http.HttpRequest) to the view’s request attribute, and any positional and/or keyword arguments [captured from the URL pattern](https://docs.djangoproject.com/en/5.1/topics/http/urls/#how-django-processes-a-request) to the args and kwargs attributes, respectively. Then [dispatch()](https://docs.djangoproject.com/en/5.1/ref/class-based-views/base/#django.views.generic.base.View.dispatch) is called.

The [`dispatch` method](https://docs.djangoproject.com/en/5.1/ref/class-based-views/base/#django.views.generic.base.View.as_view) is the method that matches the request HTTP verb to the CBV method, calls that method, and returns the resulting HttpResponse object.

Given all of that, I didn't think there were really any good hooks for users to override to customize behavior. If they override a CBV "view method" (my term for the methods with names that match HTTP verbs), they would also need to use a different URLConf than we provide in this contrib app to route requests to their customized CBV. And once you do all of that, you might as well have just created your own URLConf and function-based view yourself. FBVs are more direct and have fewer "moving parts" than Django's CBVs. Usage of this contrib app is entirely optional, and requires an additional entry in users root URLConfs anyway, so users will already need to do a bit of configuration. And quite frankly, I find Django's first-party CBVs so confusing that I use the [django-vanilla-views package](http://django-vanilla-views.org/) wherever I can instead.

I could maybe see a valid argument that CBVs can share some behavior with the CBVs in the contrib apps for other email services like SES and Sparkpost. But the webhook schema from different providers is likely to be different enough, and the webhook verification also different enough, that I don't see even CBVs that support different email services to share that much common code. But, if somebody contributes additional contrib apps to support other email services, then I think that is probably a good time to circle back around and reanalyze this decision to see if maybe there are enough similarities that it does make sense.

So I just don't see much of a benefit from switching to CBVs. IIRC, Sendgrid at least only sends event webhooks via HTTP POSTs, not any other verb (which makes sense, because presumably the webhook is triggering some sort of state change in your application so a GET request wouldn't be the appropriate verb), so allowing users to override non-`post` methods on a CBV in this contrib app doesn't really gain us anything. It might in contrib apps for other email services like SES as you note.

I mostly intend for this webhook to be used as a pretty well fleshed out example of what is possible to do with Sendgrid's event webhook. I think the view that I wrote is probably what most people will use it for if they are using this project in the first place (that's actually why I contributed it as an optional contrib app instead of releasing it as a third party project). And if somebody has a slightly different usecase that they think would be useful for others, I would be more than happy to make their behavior a configurable option.

Hope all of this makes sense. Please let me know if I should rewrite this to be a CBV anyway, or if I have convinced you.

And I'll try to fix the Ruff linting issue. :+1:

selwin:
> Hope all of this makes sense. Please let me know if I should rewrite this to be a CBV anyway, or if I have convinced you.

I think it'd have to be class based views so it can be easily reused. I haven't fully fleshed this out, but at the very least, we should have an enum that represents various events that are sent by these providers. For example:

```python
class Event(Enum):
    ACCEPTED       = "accepted"
    DELIVERED      = "delivered"
    DEFERRED       = "deferred"
    HARD_BOUNCE    = "hard_bounce"
    SOFT_BOUNCE    = "soft_bounced"
    REJECTED       = "rejected"

    # Engagement
    OPEN           = "open"
    CLICK          = "click"

    # Complaints & unsubscribes
    SPAM_COMPLAINT = "spam_complaint"
    UNSUBSCRIBED   = "unsubscribed"
```

The base class based views should have at least these methods:
* `verify_signature(request)` -> each webhook handler (Sendgrid, SES, Mailgun) would need to implement this.
* `handle_request(request, event: Event, to_addresses, subject, payload)` -> users override this for their own use.

We can then have `SendgridWebhookHandler`, `MailgunWebhookHandler`, `SESWebhookHandler` etc. But the idea is that users can simply override the `handle_request` method which already provides a whole bunch of context parsed out for them.

selwin:
Will look at this sometime in the next few days

post_office/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Event(Enum):
+    # Deliverability
+    ACCEPTED = 'accepted'
+    DELIVERED = 'delivered'
+    DEFERRED = 'deferred'
+    HARD_BOUNCE = 'hard_bounce'
+    SOFT_BOUNCE = 'soft_bounced'
```
selwin, line 27: Should be `soft_bounce` instead of `soft_bounced`, to keep it consistent with `hard_bounce`.

post_office/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+    REJECTED = 'rejected'
+
+    # Engagement
+    OPEN = 'opened'
+    CLICK = 'clicked'
```
selwin, line 32: `open` and `click`

post_office/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # Complaints & unsubscribes
+    SPAM_COMPLAINT = 'spam_complaint'
+    UNSUBSCRIBED = 'unsubscribed'
+    RESUBSCRIBED = 'resubscribed'
```
selwin, line 37: `UNSUBSCRIBE` and `RESUBSCRIBE`.

post_office/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+            )
+        else:
+            handler = self.http_method_not_allowed
+        return handler(request, *args, **kwargs)
+
+    @abstractmethod
+    def verify_webhook(self, request: HttpRequest, *args, **kwargs) -> bool:
+        pass
+
+    def process_event(self, request: HttpRequest, event: Event, email: Email | None = None, *args, **kwargs) -> None:
```
selwin, line 66: Let's keep this simple and have a single `handle_event` to start with. 

```
def handle_event(self, request, event: Event, payload: dict, *args, **kwargs):
```

`event` is whatever event that's sent and can be parsed (including `UNKNOWN`). In the future, we may also pre-parsed metadata such as `subject`, `to` etc, but would like to keep the scope simple for now.

post_office/webhook.py
blag: Fixed.

post_office/webhook.py
blag: NB: Untested since refactor.

blag: Oh, and I had to raise the minimum required Python version to 3.10 due to the `argument: Type | None = None` syntax, and only 3.10 and greater supports that.

blag:
The failing tests also fail for me when I pull the `ui/master` branch, and appear unrelated to this PR. Fixing unrelated test failures is outside the scope of this PR.

post_office/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class BaseWebhookHandler(View):
+    __metaclass__ = ABCMeta
+
+    @csrf_exempt
+    def dispatch(self, request, *args, **kwargs) -> HttpResponse:
+        # The same method as View.dispatch, but we verify the webhook once
+        # we verify the HTTP method
+        if request.method.lower() in self.http_method_names:
```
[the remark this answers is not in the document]
selwin: > But checking the HTTP method is really really cheap in terms of CPU time. Checking the webhook signature is certainly not as cheap, and may not be acceptably fast. By running the "cheap" check first, we minimize the traffic that makes it to the verify step. If we instead verify the webhook first, we spend much more CPU time verifying the webhook signature before we check the HTTP method.

You're right, I could without checking HTTP method names, but we can leave it as it is, thanks.

post_office/contrib/sendgrid/views.py
```diff
@@ -0,0 +1,296 @@
...
+            Event.SPAM_COMPLAINT,
+            Event.UNSUBSCRIBE,
+            Event.RESUBSCRIBE,
+        ]:
+            return self.log_engagement_event(request, event, payload, *args, email=email, **kwargs)
+        elif event == Event.ACCOUNT_SUSPENDED:
+            return self.account_suspended(request, event, payload, *args, email=email, **kwargs)
+        return self.unrecognized_event(request, event, payload, *args, email=email, **kwargs)
+
+    def unrecognized_event(
```
selwin, line 149: Please remove `unrecognized_event`, `log_deliverability_event` and `account_suspended`. We can cite this as an example of how to use this webhook in the docs.

post_office/contrib/sendgrid/views.py
```diff
@@ -0,0 +1,296 @@
...
+            # Raise an exception here so Django returns a 500 and emails the admins
+            # that something is wrong
+            raise Exception(
+                'Error when processing Sendgrid Events webhook. The Sendgrid Events Webhook API may have changed.'
+            )
+        return HttpResponse('ok')
+
+    def handle_event(
+        self, request: HttpRequest, event: Event, payload: dict[str, Any], *args, email: Email | None = None, **kwargs
+    ) -> None:
```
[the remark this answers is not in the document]
blag: Good call, I'll do that.

post_office/contrib/sendgrid/views.py
```diff
@@ -0,0 +1,296 @@
...
+            # Data is coming from JSON, where the root object isn't necessarily
+            # guaranteed to be a list.
+            # We reverse this because the list appears to be ordered reverse-chronologically
+            for payload in reversed(json.loads(request.body)):
+                logger.debug(payload)
+
+                # If we can't pick out the exact email, don't try to do anything
+                # This can happen if we use django-post_office and generate Emails and Logs
+                # before we configure and use this webhook handler
+                email = Email.objects.filter(message_id=payload.get('smtp-id', None)).first()
```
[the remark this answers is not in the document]
blag: Good call. I moved it into the only place where it's actually used - `log_deliverability_event`.

PR #486, opened 2025-05-19
Remove setup.py now that we have pyproject.toml

blag:
I didn't remove `setup.py` in #482 when I converted over to `pyproject.toml`. However, both files fill the same role so we should only have one source of truth. Since the Python community is moving towards `pyproject.toml` and away from `setup.py`, this PR simply removes `setup.py`, finalizing the switchover to `pyproject.toml`.


selwin:
Mind resolving the conflict so I can merge this in?

selwin:
Thanks!

PR #489, opened 2025-05-27
Add an option to set file storage

### LINKED ISSUE #382: Custom storage backend
Hi there!
Is it possible to set a specific File Storage Backend just for Post Office, instead of using Django `DEFAULT_FILE_STORAGE` setting?

blag:
Closes #382.


README.md
```diff
@@ -414,6 +414,32 @@ This section outlines all the settings and configurations that you can
...
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+    },
+    "post_office": {
+        "BACKEND": 'storages.backends.s3boto3.S3Boto3Storage',
+    },
+}
+
+POST_OFFICE = {
+    ...
+    'FILE_STORAGE': 'post_office',
```
selwin, line 438: Mind iterating on this to make declaration simpler? I'm not sure how complex this would be, but this removes one layer of indirection.

```python
POST_OFFICE = {
    'FILE_STORAGE': 'storages.backends.s3boto3.S3Boto3Storage',
}
```

blag: I respectfully disagree. I think those layers of indirection are useful to keep all database, cache, and storage settings in one single location in `settings.py`.

Do you have any specific concerns with respect to this? Is it a performance issue, or something?

There are plenty of modern third party apps (I can give examples upon request) that configure their cache to use by letting users specify the cache by key in `CACHES`, and that works pretty well. I expect it to work equally as well with `STORAGES`.

Unless you have a good reason not to, I'd like to keep it as-is.

selwin:
> I respectfully disagree. I think those layers of indirection are useful to keep all database, cache, and storage settings in one single location in settings.py.
> Do you have any specific concerns with respect to this? Is it a performance issue, or something?

Better readability, developer experience and ease of configuration. Years ago I shipped django-rq's cache configuration using the same method, [pointing to a cache backend alias](https://github.com/rq/django-rq/?tab=readme-ov-file#support-for-django-redis-and-django-redis-cache) but didn't like how it looked after it shipped.

There's always the DRY argument:

If you're setting a (new) backend for post office, why not just define it here instead of defining it elsewhere and pointing it to that other place?

vs
 
If backend is already configured elsewhere, it's easier to simply point to that other configuration.

In any case, no strong opinions here, let's keep it as it is.

PR #490, opened 2025-05-27
Add indexes to EmailTemplate

### LINKED ISSUE #250: EmailTemplate.name should be indexed
Hi there,
First of all: thank you for providing django-post-office. We are currently evaluating it as part of an internal mail-distribution service and so far it is looking good!

Templates are usually looked up by name (using `utils.get_email_template()`) but the field `EmailTemplate.name` is not indexed which may lead to bad performance.
I think it would make sense to index that field even when using a cache.

Cheers,
Konrad

blag:
Closes #250.

I added two indexes to `EmailTemplate` to help performance when querying templates.

The first index should be used by `utils.get_email_template`:

`utils.py`:
```python
    ...
    if not use_cache:
        return EmailTemplate.objects.get(name=name, language=language)
    else:
        composite_name = '%s:%s' % (name, language)
        email_template = cache.get(composite_name)

        if email_template is None:
            email_template = EmailTemplate.objects.get(name=name, language=language)
            ...
```

`models.py`:
```python
class EmailTemplate(models.Model):
    ...
    class Meta:
        ...
        indexes = [
            Index(fields=['name', 'language']),
            ...
        ]
        ...
```

The second index should be used by `EmailTemplateManager`.
`models.py`:
```python
class EmailTemplateManager(models.Manager):
    def get_by_natural_key(self, name, language, default_template):
        return self.get(name=name, language=language, default_template=default_template)

class EmailTemplate(models.Model):
    ...
    class Meta:
        ...
        indexes = [
            Index(fields=['name', 'language', 'default_template']),
            ...
        ]
        ...
```


blag:
This branch should be updated after #489 is merged, since this migration relies on that migration. That should get tests passing.

selwin:
Simply adding `db_index=True` to `name` field would suffice. `language` is a low cardinality field so it wouldn't help all that much.

PR #492, opened 2025-06-03
Fixup packaging

blag:
I didn't realize that `post_office/version.txt` was still listed in `MANIFEST.in` in PR #482.

This causes issues when attempting to install this package from a Git repository.

I also removed some of the entries in `pyproject.toml` that were causing warnings when building with `python3 -m build`, and tweaked `MANIFEST.in` to exclude test files, since we don't need to ship those.


selwin:
Thanks!

PR #493, opened 2025-07-11
feat: Add Recipient Delivery Status Fields

hery733:
Choices referenced from https://github.com/ui/django-post_office/pull/483#issuecomment-2887898269


post_office/models.py
```diff
@@ -250,6 +270,9 @@ class Log(models.Model):
     )
     date = models.DateTimeField(auto_now_add=True)
     status = models.PositiveSmallIntegerField(_('Status'), choices=STATUS_CHOICES)
+    recipient_delivery_status = models.PositiveSmallIntegerField(
+        _('Recipient Delivery Status'), choices=RecipientDeliveryStatus.choices, blank=True, null=True
+    )
```
selwin, line 275: These two fields can be combined into `status` field. A log can only have one status.

post_office/models.py
```diff
@@ -27,6 +27,23 @@
 STATUS = namedtuple('STATUS', 'sent failed queued requeued')._make(range(4))
 
 
+class RecipientDeliveryStatus(models.IntegerChoices):
```
selwin, line 30: Can we group the positive and negative cases together?

```python
ACCEPTED = 10, _('Accepted')
DELIVERED = 20, _('Delivered')
OPENED = 70, _('Opened')
CLICKED = 80, _('Clicked')

DEFERRED = 30, _('Deferred')
SOFT_BOUNCED = 40, _('Soft Bounced')
HARD_BOUNCED = 50, _('Hard Bounced')

SPAM_COMPLAINT = 90, _('Spam Complaint')
UNSUBSCRIBED = 100, _('Unsubscribed')
```

I also took out `REJECTED` because this is a response that's returned by ESPs and not by recipient's mail server. Perhaps this could be added to the `status` field? But this should be done in a separate PR.

PR #498, opened 2025-08-01
add jsonfield migration

### LINKED ISSUE #497: Error when opening email details
the error is displayed as below
```
UndefinedFunction: operator does not exist: text -> unknown
LINE 1: ...1194 AND NOT (("post_office_attachment"."headers" -> 'Conten...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.

  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
ProgrammingError: operator does not exist: text -> unknown
LINE 1: ...1194 AND NOT (("post_office_attachment"."headers" -> 'Conten...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts
```

Image

upon checking the headers is not actually a jsonfield, yet. will make PR to migrate the fields to jsonfield

marsha97:
Fix #497 

project which use post office before jsonfield migration will encounters the problem. since it was not really migrated into jsonfield before. so it stays as text(refer to issue)

after migration is done, the field is changed into jsonb
image

query can be performed without error now
image

the old json data before migration also still intact. and still queryable


marsha97:
will do migration on db lvl instead for now

```
ALTER TABLE "post_office_attachment" ALTER COLUMN "headers" TYPE jsonb USING "headers"::jsonb;
```

PR #516, opened 2026-01-12
Add fallback language support in get_template with try/catch error ha…

### LINKED ISSUE #515: Support fallback language in get_template with try/catch error handling
## Description
Currently, in the `get_template` method in `post_office/mail.py` (around line 200), if a translation lookup fails, the entire operation may fail or throw an exception. 

## Proposed Solution
We should wrap the template retrieval logic in a try/catch block so that if we don't have a translation for the requested language, we gracefully fall back to the original email template instead of failing completely.

## File Reference
https://github.com/ui/django-post_office/blob/99be3b50a904d84f35fecb14f6c64575611a2e8a/post_office/mail.py#L200

## Benefits
- Improved robustness and reliability of the email system
- Graceful degradation when translations are unavailable
- Better user experience by always sending emails, even if translation fails
- Prevents application crashes due to missing language translations

yakov-sudo:
…ndling

Fixes #515

This PR wraps template retrieval logic in try/catch blocks so that:
- If a translated template lookup fails via translated_templates.get(), the original template is kept
- If get_email_template() fails for a specific language, it falls back to getting the template without language

This prevents the entire email sending operation from failing when a translation is unavailable.

### DIFF
--- a/post_office/mail.py
+++ b/post_office/mail.py
@@ -201,9 +201,16 @@ def send(
             template = template
             # If language is specified, ensure template uses the right language
             if language and template.language != language:
-                template = template.translated_templates.get(language=language)
+                                    try:
+                                                                template = template.translated_templates.get(language=language)
+                                    except EmailTemplate.DoesNotExist:
+                                                                pass  # Keep original template if translation not found
         else:
-            template = get_email_template(template, language)
+            try:
+                                template = get_email_template(template, language)
+            except EmailTemplate.DoesNotExist:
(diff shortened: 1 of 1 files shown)


selwin:
Hey, sorry but I actually don't quite agree with this change. I think the current behavior of raising an exception is correct so the exception can be handled by the user. If you feel strongly, we can add an argument that alters the behavior such that it returns the original template if the translation is not found.

PR #521, opened 2026-04-15
Separate backend each threads on send_queued_mail

ibadarrohman:
## LLM Output Claude Sonnet 4.6

**Root Cause:** In `_send_bulk`, `prepare_email_message()` is called in the **main thread** before spawning worker threads. This embeds the main thread's `AnymailRequestsBackend` connection (and its `requests.Session`) into every email's `_cached_email_message`. When the `ThreadPool` workers dispatch those emails, all threads concurrently call `session.request()` on the **same shared `requests.Session`**, which is [explicitly not thread-safe](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects). This corrupts socket state, causing `[Errno 11] Resource temporarily unavailable` at the OS level.

The fix: in `_send_email`, replace the pre-cached connection with a thread-local one from `connections[]` before dispatching. Since `ConnectionHandler` uses `threading.local`, each worker thread will get its own `AnymailRequestsBackend` with its own `requests.Session`. 

## Deeper Analysis
---

## 1. Why `requests.Session` is not thread-safe

A `Session` object holds mutable shared state:

```
session.cookies      ← mutable CookieJar
session.headers      ← mutable dict
session.adapters     ← HTTPAdapter with its PoolManager
session.env_proxies  ← mutated on each request()
```

The critical one is `session.request()` itself — it performs a **non-atomic read-modify-write** sequence:

```python
# Inside requests/sessions.py (simplified)
def request(self, method, url, ...):
    req = Request(...)
    prep = self.prepare_request(req)   # reads self.headers, self.cookies
    settings = self.merge_environment_settings(...)  # reads+writes self.env_proxies
    return self.send(prep, **settings)
```

When Thread A and Thread B both call `session.request()` at the same time, they both enter `prepare_request()` and `merge_environment_settings()` concurrently — reading and writing the same dict/cookie objects without locks. This is a classic TOCTOU (Time-of-Check-Time-of-Use) race.

---

## 2. Why does sharing corrupt socket state?

Go one level deeper: `HTTPAdapter` owns a `urllib3.PoolManager`, which maps hostnames → `HTTPConnectionPool`. Each pool has a `queue.LifoQueue` of `HTTPConnection` objects.

When 5 threads share the **same** adapter/pool and all call `adapter.send()` simultaneously:

```
Thread 1 ──→ pool.urlopen() → get conn from queue → send bytes → read response
Thread 2 ──→ pool.urlopen() → get conn from queue → ← queue is empty, create new conn
Thread 3 ──→ pool.urlopen() → get conn from queue → ← queue is empty, create new conn
Thread 4 ──→ pool.urlopen() → get conn from queue → ← queue is empty, create new conn
Thread 5 ──→ pool.urlopen() → get conn from queue → ← queue is empty, create new conn
```

All 5 threads simultaneously call `HTTPConnection._new_conn()` → `socket.create_connection()` → internally calls `socket.getaddrinfo()` for DNS.

At the OS level, DNS resolution requires creating its **own** temporary UDP socket to query the resolver. Five threads doing this at exactly the same instant means the OS must allocate 5 resolver sockets simultaneously, on top of the 5 TCP sockets being opened. When system resources are tight (file descriptor pressure, socket buffer limits, or the resolver queue is full), the kernel returns **EAGAIN** on one of those socket allocations.

---

## 3. Why specifically Errno 11 (`EAGAIN` / `Resource temporarily unavailable`)

`EAGAIN` (errno 11 on Linux) means: **"I can't do this right now — try again later."** It's the kernel's way of saying a resource is momentarily exhausted without hard-failing.

The exact call that fails, from the traceback:

```python
# urllib3/util/connection.py
for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
```

`getaddrinfo` creates a non-blocking DNS socket internally. Under the hood, libc's resolver does:

```c
// Creates a UDP socket to talk to /etc/resolv.conf nameserver
fd = socket(AF_INET, SOCK_DGRAM | SOCK_NONBLOCK, ...)
// If the system socket table is temporarily saturated:
// → returns -1, errno = EAGAIN
```

So the chain is:

```
5 threads × (1 TCP socket + 1 DNS socket) = 10 simultaneous socket() syscalls
         ↓
system socket buffer / fd slots temporarily saturated
         ↓
socket() syscall returns EAGAIN
         ↓
Python raises BlockingIOError: [Errno 11]
         ↓
urllib3 wraps it as NewConnectionError
         ↓
requests wraps it as ConnectionError
         ↓
anymail wraps it as AnymailRequestsAPIError
```

---

## Why per-thread sessions fix it

With the fix, each worker thread calls `connections[alias]` which hits `threading.local` storage. The first access on a given thread creates a **brand new** `AnymailRequestsBackend` → new `requests.Session` → new `HTTPAdapter` → new `PoolManager`. Now:

```
Thread 1 → its own Session → its own socket lifecycle
Thread 2 → its own Session → its own socket lifecycle
Thread 3 → its own Session → its own socket lifecycle
```

No shared state, no concurrent mutation, sockets are created sequentially within each thread's own connection lifecycle — no `EAGAIN`.

### DIFF
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -104,7 +104,7 @@ class Email(models.Model):
         self._cached_email_message = None
 
     def __repr__(self):
-        return '<%s: %s>' % (self.__class__.__name__, self.to)
+        return f'<{self.__class__.__name__}: {self.to}>'
 
     def __str__(self):
         return f'{", ".join(self.to)}'
@@ -139,6 +139,7 @@ class Email(models.Model):
             multipart_template = None
             html_message = self.html_message
 
+        connection = connections[self.backend_alias or 'default']
         if isinstance(self.headers, dict) or self.expires_at or self.message_id:
             headers = dict(self.headers or {})
             if self.expires_at:
@@ -158,6 +159,7 @@ class Email(models.Model):
                     bcc=self.bcc,
                     cc=self.cc,
                     headers=headers,
+                    connection=connection,
                 )
                 msg.attach_alternative(html_message, 'text/html')
             else:
@@ -169,6 +171,7 @@ class Email(models.Model):
                     bcc=self.bcc,
                     cc=self.cc,
                     headers=headers,
+                    connection=connection,
                 )
                 msg.content_subtype = 'html'
             if hasattr(multipart_template, 'attach_related'):
@@ -183,6 +186,7 @@ class Email(models.Model):
                 bcc=self.bcc,
                 cc=self.cc,
                 headers=headers,
+                connection=connection,
             )
 
         for attachment in self.attachments.all():
@@ -325,7 +329,7 @@ class EmailTemplate(models.Model):
         ordering = ['name']
 
     def __str__(self):
-        return '%s %s' % (self.name, self.language)
+        return f'{self.name} {self.language}'
 
     def natural_key(self):
         return (self.name, self.language, self.default_template)
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -217,7 +217,7 @@ class Email(models.Model):
         """
         try:
             msg = self.email_message()
-            if connection:
+            if connection is not None:
                 msg.connection = connection
             msg.send()
             status = STATUS.sent
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -112,8 +112,14 @@ class Email(models.Model):
     def email_message(self):
         """
         Returns Django EmailMessage object for sending.
+
+        The connection is always re-fetched from the thread-local registry so
+        that worker threads each use their own backend/session rather than
+        sharing the one that was embedded when prepare_email_message() ran in
+        the main thread.
         """
         if self._cached_email_message:
+            self._cached_email_message.connection = connections[self.backend_alias or 'default']
             return self._cached_email_message
 
         return self.prepare_email_message()
--- a/post_office/models.py
+++ b/post_office/models.py
@@ -209,12 +209,16 @@ class Email(models.Model):
     def dispatch(self, log_level=None, disconnect_after_delivery=True, commit=True, connection=None):
         """
         Sends email and log the result.
+
+        If ``connection`` is provided, it overrides the connection embedded in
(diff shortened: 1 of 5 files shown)


post_office/models.py
```diff
@@ -112,8 +112,14 @@ def __str__(self):
...
         """
         Returns Django EmailMessage object for sending.
+
+        The connection is always re-fetched from the thread-local registry so
+        that worker threads each use their own backend/session rather than
+        sharing the one that was embedded when prepare_email_message() ran in
+        the main thread.
         """
         if self._cached_email_message:
+            self._cached_email_message.connection = connections[self.backend_alias or 'default']
```
selwin: Can we do this instead so it's more explicit?

Image

Modify `dispatch()` to accept an optional `connection` argument. I think this approach is cleaner and less magical.

tests/test_mail.py
```diff
@@ -140,31 +131,37 @@ def test_send_bulk(self):
...
+            },
+            'THREADS_PER_PROCESS': 1,
+        },
+    )
     def test_send_bulk_reuses_open_connection(self):
         """
-        Ensure _send_bulk() only opens connection once to send multiple emails.
+        Ensure _send_bulk() opens one connection per thread, not one per email.
+        With THREADS_PER_PROCESS=1: main thread opens one during prepare, worker
+        thread opens one during send — total 2 opens for any number of emails.
```
ibadarrohman, line 147: FYI @selwin

post_office/models.py
```diff
@@ -206,12 +206,20 @@ def prepare_email_message(self):
...
+
+        If ``connection`` is provided, it overrides the connection embedded in
+        the email message by ``prepare_email_message()``. This allows callers
+        (e.g. worker threads) to supply a thread-local connection rather than
+        reusing one that was opened in a different thread.
         """
         try:
-            self.email_message().send()
+            msg = self.email_message()
+            if connection:
```
selwin: Can you change this to `if connection is not None:`?

PR #523, opened 2026-06-15
Close worker-thread connections opened in _send_bulk

jorikbraet:
## Problem

PR #521 (released in 3.11.2) moved the email-backend connection open into the ThreadPool workers inside _send_bulk:

    def _send_email(email, log_level):
        connection = connections[email.backend_alias or 'default']
        email.dispatch(..., connection=connection)

But ConnectionHandler stored its registry on a threading.local, so the connections.close() called from the main thread in _send_bulk's finally only saw the main thread's local cache. SMTP connections opened on the worker threads were never .quit()-ed. They survived until garbage collection, producing:

    ResourceWarning: unclosed <ssl.SSLSocket ...>

and leaking file descriptors in long-running send_queued_mail workers.

## Fix

Switch ConnectionHandler from threading.local to a process-wide dict keyed by (threading.get_ident(), alias) under a lock. Per-thread isolation is preserved (each thread still gets its own connection, no sharing or contention), and close() called from any thread now sees and closes every worker's connection.

This preserves the intra-batch connection reuse from #521. _send_email running multiple times on the same worker thread still hits the cache rather than reopening. The existing test_send_bulk_reuses_open_connection still passes at 2 opens for 3 emails (1 main + 1 worker, reused).

I considered the smaller fix of wrapping _send_email's body in a try/finally that calls connections.close() per call. That works but defeats #521's reuse: every email pays a fresh open() (plus STARTTLS, for SMTP). The registry change is the same number of files touched and keeps the perf shape #521 was after, so it seemed the right call. Happy to switch to the per-call close if you prefer the smaller surface.

## Test

Added test_send_bulk_closes_worker_connections: a thread-safe counting backend tracks open() and close() calls; the test runs _send_bulk over 5 emails with THREADS_PER_PROCESS=2 and asserts opens == closes.

- Pre-fix: fails (opens=2, closes=1, the worker's open is never closed).
- Post-fix: passes.

Full test suite still green locally on Django 5.2 (django-admin test post_office ./).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+## Unreleased
+* Fixed a bug where SMTP connections opened on `_send_bulk` worker threads were never closed (causing `ResourceWarning: unclosed <ssl.SSLSocket ...>` leaks in production). `ConnectionHandler` now tracks connections across threads so `close()` from the main thread reaches every worker's connection.
+
 ## Version 3.11.2 (2026-04-16)
 * `_send_email` should dispatch email using it's own connection. Thanks @ibadarrohman!
 
--- a/post_office/connections.py
+++ b/post_office/connections.py
@@ -1,4 +1,4 @@
-from threading import local
+import threading
 
 from django.core.mail import get_connection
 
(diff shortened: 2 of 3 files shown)


selwin:
Thanks for the PR, I'm merging in https://github.com/ui/django-post_office/pull/526 because it's a bit easier to understand and doesn't use `threading.Lock`

Issue #457, opened 2023-12-26
Is this a correct field rendering in email template?


heckad:
![image](https://github.com/ui/django-post_office/assets/45011689/c3c36166-4701-4b76-8501-1bae3490b5f6)
Why do the fields look so bad? Is it a wrong setting or something else?

selwin:
Fixed in https://github.com/ui/django-post_office/pull/524

heckad:
Wow, cool!

Issue #455, opened 2023-11-27
insanely hard to edit default content for a template


kakulukia:
Hi 

![SCR-20231127-jd1](https://github.com/ui/django-post_office/assets/96352/08aa9cc2-5f0c-4010-a1e4-73742db01ddd)

I just discovered that is **IS** actually possible to edit the default content, BUT really you have to know that and where to click for editing the content. In that screenshot i have marked the very tiny area with a custom border for the subject.

You only see the span without any indication of its size. I would expect a normal text input. Why is it that complicated?
I first thought, i have to add multiple language templates and only then can mark one as the default, but no .. that interface should be optimised!

I will add CSS to my admin to at least display a border for ".CodeMirror pre".
But this should be fixed in some future version as this admin interface is not usable that way!

selwin:
@kakulukia PR to fix this welcome :)

kakulukia:
I would remove all the pre stuff and just display the normal admin fields, but why was that overlay added in the first place?
It might have been a good idea, but i dont get it yet.

kakulukia:
Okay, after playing around with it a bit more i see that the current elements display code highlights. Thats nice, but still the borders are missing to actually know where to click to start entering text.

selwin:
Fixed in https://github.com/ui/django-post_office/pull/524

PR #524, opened 2026-06-19
Enhance CodeMirror styling in email template change form.

c-kaiser-bremer:
After struggeling to see the input fields for the standard mail templates and seeing #455 and #457 I wanted to add a simple solution.
I added a minimum width to the fields and added the same border as the other Admin input fields.
Now the inputs are clearly visible in the template form and are similar in style to the other Admin inputs.

| Before | After |
|--------|--------|
| Bildschirmfoto 2026-06-19 um 14 07 46 | Bildschirmfoto 2026-06-19 um 14 07 32 |


selwin:
Thanks!

PR #526, opened 2026-06-30
Fix: Close SMTP connections opened in worker threads in _send_bulk()

simonphilips:
When sending queued mail, `_send_bulk` runs each send in a `ThreadPool` worker that opens its own connection in thread-local storage. The main thread's `connections.close()` can't access those, so the sockets were leaked and only cleaned up at garbage collection (which triggered a lot of warnings for us). 

To solve this, collect the worker connections and close them once the pool has finished. This keeps re-using one connection per worker thread for the whole batch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+## Master
+* Fixed a connection leak in `_send_bulk()`.
+
 ## Version 3.11.2 (2026-04-16)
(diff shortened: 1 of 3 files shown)


simonphilips:
I just missed that there's already another PR for it but it's a bit cryptic and does some things that aren't needed (like the locking, dicts are thread-safe even without the GIL).

selwin:
Thanks!

